PR #11, opened 2016-05-10
LegendChat support.

Time6628:


Scarsz:
Make the pull request target the `develop` branch & form the hook around how the Herochat hook integrates

Time6628:
I will do so when I can, :cat:.

PR #25, opened 2016-08-02
Resolve log4j-dependency

Androkai:
log4j-core version 2.0-beta9 was the needed version. Works fine now: [https://jenkins.androkai.net/job/DiscordSRV-test/3/](https://jenkins.androkai.net/job/DiscordSRV-test/3/)


Scarsz:
Leave the old log4j entry there, comment it out. Never know if we might need it again. Will merge after

Androkai:
Done

PR #59, opened 2016-09-22
Adding LunaChat co-operation.

ucchyocean:
Regarding the following comment.
https://dev.bukkit.org/bukkit-plugins/discordsrv/#c49

Please review the source codes, and merge for DiscordSRV v12.0.
Thanks!

<s>
NOTE: Regarding the default value of "LunaChatHook" configuration, "false" is better I think.
Because, most of Japanese servers are using LunaChat only for chat coloring and IME (Kana-Kanji conversion).
A handful of people, are using the channel chat feature of LunaChat. I will announce how to connect channel chat with Discord to them.
</s>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,10 @@
             <id>scarsz</id>
             <url>https://dl.bintray.com/scarsz/maven</url>
         </repository>
+        <repository>
+            <id>LunaChat</id>
+            <url>https://raw.github.com/ucchyocean/mvn-repo/master</url>
+        </repository>
     </repositories>
 
     <dependencies>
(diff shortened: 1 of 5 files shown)


ucchyocean:
Hm...
OK, I follow your convention.

Androkai:
Hey @ucchyocean nice to hear from you. Because @Scarsz seems to be busy I merge your commit now. I switched the hook-option back to "true" by default to make DiscordSRV work out of the box (Is a to big problem for some users to set it true :wink:) with LunaChat.

Great thx for the implementation.

ucchyocean:
Thank you so much! 👍

PR #60, opened 2016-09-23
Escape markdown before sending messages to Discord.

TheReverend403:
Prevents usernames such as `_username_` being send as italicised text.


TheReverend403:
Woops, wrong branch.

PR #78, opened 2016-10-10
Add discordsrv.admin permission

TheReverend403:


Androkai:
Thx for the request and sorry for the late reply. Because of the core from DiscordSRV is being rewritten by Scarsz for the upcoming [v0.13](https://github.com/Scarsz/DiscordSRV/tree/13.0) we freeze all bigger changes on the code at the moment. Scarsz will decide about then this is done.

PR #168, opened 2017-04-08
Continue loop only when not subscribed.

cnaude:


Scarsz:
Be sure to develop against `develop` next time; thanks for the PR

PR #172, opened 2017-04-13
Updated Japanese translation messages.

ucchyocean:
Hi.
I have updated the Japanese translation messages.
Please review and merge these codes.


Scarsz:
You should probably join our Discord server by the way to get notified when new strings are added for translating

ucchyocean:
OK, I will join when I get home.

PR #183, opened 2017-05-18
Do not use event incoming chat state

FabioZumbi12:


Scarsz:
I still don't like how we're working around an issue instead of resolving it.

PR #195, opened 2017-06-05
Strip colors in player name when using Webhook

Rodrigo48:
This change prevents color codes such as "§6" appearing with the username.


Scarsz:
Change it to DiscordUtil's stripColor & I'll merge it

PR #198, opened 2017-06-09
Fix up DebugListener

AustinLMayes:
Instead of hardcoding developer IDs, they're in the config now. Also, the old if-statement didn't take into account PMing the bot to abuse the eval functionality. The source of the message really doesn't matter anyway


Androkai:
@Scarsz has to decide about this but here is a personal statement:

I agree with your suggestion to make it possible to remove the developer IDs easily via config option but your actual solution completely remove use from the default configuration which miss the goal of making us able to help server owners with heavy problems to setup DiscordSRV which also often mess up their config etc.

If we approach this topic now I suggest the use of a second configuration file within the plugin jar which is not written to the file system by default and contain hidden config options which are not present in the normal one but make advanced paranoid users able to remove us from the list by adding the option manually to the config.yml and changing the setting. This can be achieved by using an inheritance system which get default values from the hidden file if not present in the normal one. This also provides that all hidden options are documented in one place and can easily be changed. One problem I see is that this means that we also have to change something in the config migration code.

Scarsz:
Couple things wrong with this...

1. You developed against `master`
2. The changes for the if statement would break it, not improve it
3. The debuggers list doesn't need to be a config option, me & @Androkai are likely the only ones who'll ever be in that list

AustinLMayes:
@Scarsz Couple things...

1. If you look at the changes, nothing if even effected by this, just squash and merge.
2. IF not on authorized list OR IF not guild owner, DENY. Seems fine to me. The old if statement was broken in the first place.
3. Then remove the eval functionality, since you have effectible wirrteen a backdoor in your plugin which is against spigot rules.
4. Had to add the config option.

Scarsz:
The only point you've really made is #3. I can see the security concern about that and I'll change it, but it should clearly be noted that it's a debugging thing. The command wouldn't even work on 99% of servers anywho because Nashorn isn't included in the JRE. Either way, will change that.

And no, the if statement isn't broken.

AustinLMayes:
@Scarsz Let's look this over then

```
 if (!authorized.contains(event.getAuthor().getId()) && // not a dev
                 (event.isFromType(ChannelType.TEXT) && // Not in channel
!event.getGuild().getOwner().getUser().getId().equals(event.getAuthor().getId())) // Not a guid owner
```

Will fail to return if:
1. Bot is PMed from a single person
2. Bot is PMed from a group DM.

All if effectively did was remove the channel check because as I said in the original PR, it is not needed.

Scarsz:
I'm starting to see what you're trying to say, but your if statement is broken as well. If the bot gets PM'd, the Event#getGuild returns null

AustinLMayes:
@Scarsz Here's the fixed version
```
 if (!DiscordSRV.getPlugin().getAuthorizedDebuggers().contains(event.getAuthor().getId()) ||
                (event.getGuild() != null && !event.getGuild().getOwner().getUser().getId().equals(event.getAuthor().getId())
        ) return;
```

Also on the Spigot page just document that if you need help from a developer, add their ID to the config.

Scarsz:
This "authorized debuggers" thing isn't going to exist. The only changes I'm implementing from this are the if statement change and restricting the eval command to only being used in my guild.

shadowolfyt:
Here you go, I have provided a fixed version. I am a well known user of your plugin and Austin showed me this bug. through being a user on my server he gave me all of this [https://gist.github.com/anonymous/96d6507de3d478392956324387777ca8] and he told me this file fixed that issue.
All he's done is the following: Change ConfigVersion to ConfigVersion: 14.1 and add DebuggerIds: [] somewhere in the file

"The only changes I'm implementing from this are the if statement change and restricting the eval command to only being used in my guild."
You should think of other servers too not just your own.
[DiscordSRV-Build-14.4-SNAPSHOT.zip](https://github.com/Scarsz/DiscordSRV/files/1065542/DiscordSRV-Build-14.4-SNAPSHOT.zip)

PR #206, opened 2017-07-15
Replaced while + sleep with blocking operation

Almighty-Alpaca:


Almighty-Alpaca:
why :sob:

Scarsz:
I specifically don't want to use `JDABuilder#buildBlocking` & I already changed the JDA#shutdown call myself in a later commit

Almighty-Alpaca:
`JDABuilder#buildBlocking()` is made for exactly that purpose

Scarsz:
Okay, but notice me doing it with buildAsync gives me the JDA object while it's being built whereas buildBlocking doesn't give it until after it's completely loaded. If the debug handler tries to do a report for DiscordSRV and JDA isn't built yet, not only will it error on multiple things but it will be unable to debug JDA building issues.

PR #221, opened 2017-08-05
Added support for PhantomAdmin vanish.

soliddanii:
Simple. Added the hook for PhantomAdmin plugin and the condition to check if it is enabled. It works, i'm using it. Thank you.


Scarsz:
Please read https://github.com/Scarsz/DiscordSRV/blob/master/CONTRIBUTING.md next time

PR #294, opened 2018-01-20
NL Translation

tomkeuper:
Added Config and Messages Dutch translations.


Scarsz:
https://github.com/tomkeuper/DiscordSRV/blob/patch-1/src/main/java/github/scarsz/discordsrv/util/LangUtil.java needs to be translated as well

PR #312, opened 2018-02-21
Fix for TownyChat 0.52 version hook

huynhqtienvtag:
I had a error with the TownyChat hook. And I recognized the changes from TownyChat.
This is a small fix that makes DiscordSRV work with newer versions of TownyChat.


huynhqtienvtag:
They have changed TownyChat.java back to Chat.java
So this pull request is not working for the recent dev-build.

PR #354, opened 2018-06-23
A way to keep roles on your discord even after you loose perms in Minecraft

trent304:
This would be a nice feature to have so that I could use it in the event of a reset to know who had what rank.


Scarsz:
Do you not realize that you made a pull request?

PR #365, opened 2018-07-21
Changed code generation logic

sbienkow:
In the previous code, there was 10% chance it'll loop again.
By introducing my changes, there's `10000 / linkingCodes.size()` % chance to loop again, but it also removes the possibility of overwriting currently existing codes.
It could be improved by generating a set of codes that aren't used and then randomly selecting one. Especially with max 10000 entries this shouldn't add much overhead.

It also allows link codes to start with zero, i.e. "0000" and "0001" will be valid codes.


Scarsz:
While I don't really agree with the old logic being "inefficient" per se, this implementation does handle collisions better so I'll accept it

PR #382, opened 2018-08-25
fix DoNotSendPhrases condition in whitelist cases

### LINKED ISSUE #381: doNotSendActsAsWhitelist not works properly as expected
https://github.com/DiscordSRV/DiscordSRV/blob/79dc7d6d17f1df41e64a7edd527705e3c8049868/src/main/java/github/scarsz/discordsrv/objects/log4j/ConsoleAppender.java#L115-L117

In whitelist case, it means the `line` need to contain **all** phrases of the array.

Example:
```java
Boolean doNotSendActsAsWhitelist = true;
String line = "shikendon lost connection: Timed out";
String[] phrases = new String[]{"Timed out", "Whatever"};

for (String phrase : phrases)
    if (line.contains(phrase) == !doNotSendActsAsWhitelist) return;
System.out.println(line); // Unreachable
```

Expected:
```java
// import java.util.Arrays;

Boolean doNotSendActsAsWhitelist = true;
String line = "shikendon lost connection: Timed out";
String[] phrases = new String[]{"Timed out", "Whatever"};

if (Arrays.stream(phrases).parallel().anyMatch(line::contains) == !doNotSendActsAsWhitelist) return;

System.out.println(line); // shikendon lost connection: Timed out
```

Message should be send when `line` contains any phrases of whitelist.

shikendon:
Try to fix #381 

I don't have Java environment in this laptop. But this changes should works. Thanks.

~~Opps, wait a minute... found a typo on it.~~ Fixed

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/objects/log4j/ConsoleAppender.java
+++ b/src/main/java/github/scarsz/discordsrv/objects/log4j/ConsoleAppender.java
@@ -32,6 +32,7 @@ import org.apache.logging.log4j.core.layout.PatternLayout;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.List;
 
 @SuppressWarnings("unchecked")
 @Plugin(name = "DiscordSRV-ConsoleChannel", category = "Core", elementType = "appender", printObject = true)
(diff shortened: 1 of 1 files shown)


Scarsz:
Is stream parallelism really necessary for this?

shikendon:
`parallel()` is not necessary. If there are any concerns, it can be removed directly.

Scarsz:
`parallel` would cause negative performance. It takes a lot of CPU time to get the stream in parallel mode compared to just running sequentially. The list would only have a couple entries, too.

VTCAKAVSMoACE:
There is no reason to use `parallel()` here at all, since it's all handled in one thread.

shikendon:
Thanks for reviewing. I removed it.

mibby:
```
COMPILATION ERROR : 
-------------------------------------------------------------
github/scarsz/discordsrv/objects/log4j/ConsoleAppender.java:[117,61] incompatible types: java.util.List<java.lang.String> cannot be converted to java.lang.String[]
1 error
```

shikendon:
@mibby Thanks. Fixed and tested on my server.

PR #429, opened 2019-02-09
Achievements > advancements

Madis0:
Minecraft 1.12+ uses _advancements_ instead of _achievements_, therefore the messages in DiscordSRV should be updated too. 

I did not change any underlying code, as I feel this is not worthy of breaking every admin's config.

### DIFF
--- a/src/main/resources/messages/ru.yml
+++ b/src/main/resources/messages/ru.yml
@@ -147,7 +147,7 @@ MinecraftPlayerDeathMessageFormat: ":skull: **%deathmessage%**"
 # %date%:        текущие время и дата
 # Также поддерживаются PlaceholderAPI шаблоны
 #
-MinecraftPlayerAchievementMessagesFormat: ":medal: **%displayname% только что получил достижение %achievement%!**"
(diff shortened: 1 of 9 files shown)


Vankka:
Ideally all languages should be updated, if we're changing one

Madis0:
I updated all currently translated messages from the latest completed ones in Crowdin.

src/main/resources/messages/ru.yml
```diff
@@ -147,7 +147,7 @@ MinecraftPlayerDeathMessageFormat: ":skull: **%deathmessage%**"
 # %date%:        текущие время и дата
 # Также поддерживаются PlaceholderAPI шаблоны
 #
-MinecraftPlayerAchievementMessagesFormat: ":medal: **%displayname% только что получил достижение %achievement%!**"
+MinecraftPlayerAchievementMessagesFormat: ":medal: **%displayname% получил достижение «%achievement%»!**"
```
Vankka: « » shouldn't be around %achievement%

Madis0: Was on the original, but fair enough.

Madis0:
Based this upon the wrong branch (master instead of develop), continued on #430

PR #430, opened 2019-02-09
Achievements > advancements

Madis0:
Minecraft 1.12+ uses _advancements_ instead of _achievements_, therefore the messages in DiscordSRV should be updated too. 

I did not change any underlying code, as I feel this is not worthy of breaking every admin's config.

### DIFF
--- a/src/main/resources/config/en.yml
+++ b/src/main/resources/config/en.yml
@@ -6,7 +6,7 @@ BotToken: "BOTTOKEN"
 
 # Channel links from game to Discord
 # syntax is Channels: {"in-game channel name": "numerical channel ID from Discord", "another in-game channel name": "another numerical channel ID from Discord"}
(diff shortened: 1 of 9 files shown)


Scarsz:
I kept it as `achievement` because that's more explanatory than `advancement`

Madis0:
But it differs on what the user sees in game...

Vankka:
Yeah & it will keep differing: just for <1.12 users, there are people still using version even down to 1.7.10

Scarsz:
Someone *completely* new to Minecraft will know what an achievement is, not everyone would know what an advancement is. It's just a case about explaining the most information to the most amount of people.

Madis0:
How about changing the strings only?

Scarsz:
I’d prefer something more like `advancement/achievement`, really. Doesn’t look the best but I guess it’s the best of both worlds.

Madis0:
Changed the messages yml to reflect that.

PR #469, opened 2019-05-22
chinese translation

kizajan:
add the translation of traditional Chinese

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/util/LangUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/LangUtil.java
@@ -55,7 +55,8 @@ public class LangUtil {
         NL("Dutch"),
         ES("Spanish"),
         RU("Russian"),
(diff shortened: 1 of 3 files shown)


Scarsz:
How do you get around the GFW's block of Discord? I don't know how useful this would be to merge since nobody's able to use Discord in China without VPNs/proxies anyway.

kizajan:
dude...lots of people speak chinese , not only in china

Scarsz:
Clearly people speak Chinese outside of China... 😐

My point is that how many people are actually going to be using this translation when Discord itself is banned in China? The primary population that speaks Chinese can't use the service that this plugin is made to use.

kizajan:
idn
maybe more than the estonian speakers

Vankka:
There are a few countries south of China that do have a lot of Chinese speakers, indeed with higher population than Estonia

PR #479, opened 2019-05-27
Implement MySQL account link backend

Scarsz:


Vankka:
https://github.com/DiscordSRV/DiscordSRV/blob/3ce3cd33804ecd13bb3da45d019d2602dc6314cf/src/main/java/github/scarsz/discordsrv/util/ConfigUtil.java#L136

kuohsuanlo:
mysql-account-backend seems using sync thread. In a server with around 60000 offline players, 9000 discord members, any player that attempts to use /discord link would freeze the entire server. Is it possible to make the link associated function async? 

Using file-based-backend does not cause such issue.

PR #494, opened 2019-06-30
Fix #485

### LINKED ISSUE #485: Properly blocking @everyone and @here
**Bug report**
It turns out that it's still possible to abuse the bot to ping `@everyone`.

**Steps to reproduce**
1) Type `Hello @@everyoneeveryone`
2) Everyone will get pinged

The reason for that is because `Hello @@everyoneeveryone` contains `@everyone`. This part gets replaced with *nothing*. So, it turns into `Hello @everyone`.

A quick fix would be to replace it with a single space instead of nothing. Then `Hello @@everyoneeveryone` would turn into `Hello @ everyone`.

**Relevant source code**
https://github.com/DiscordSRV/DiscordSRV/blob/3ce3cd33804ecd13bb3da45d019d2602dc6314cf/src/main/java/github/scarsz/discordsrv/util/DiscordUtil.java#L239

stefnotch:
Fix #485 which is a major issue as soon as players figure out how to abuse it.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/util/DiscordUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/DiscordUtil.java
@@ -236,7 +236,7 @@ public class DiscordUtil {
                 String before = message;
                 for (String phrase : DiscordSRV.config().getStringList("DiscordChatChannelCutPhrases")) {
                     // case insensitive String#replace(phrase, "")
-                    message = message.replaceAll("(?i)" + Pattern.quote(phrase), "");
+                    message = message.replaceAll("(?i)" + Pattern.quote(phrase), " ");
                     changes = before.length() - message.length();
                 }
             } while (changes > 0); // keep cutting until there were no changes


Scarsz:
1. You've made your request to `master`, which isn't allowed
2. This isn't the proper fix. The actual issue is a logic issue with the changes number being updated when it shouldn't be. I just haven't had the time to fix it recently.

stefnotch:
@Scarsz My bad, I should have checked out [Contributing.md](https://github.com/DiscordSRV/DiscordSRV/blob/master/CONTRIBUTING.md) before doing this PR. 

> This isn't the proper fix. The actual issue is a logic issue with the changes number being updated when it shouldn't be. I just haven't had the time to fix it recently.

So, it does run the `replace` code repeatedly until nothing changes anymore?

Scarsz:
> So, it does run the `replace` code repeatedly until nothing changes anymore?

No, it doesn't. That's what it **should** do. The changes value is being changed after every blocked phrase instead of accumulating a total from all of the phrases during the loop.

stefnotch:
@Scarsz Ok, that clears stuff up.

PR #577, opened 2019-11-23
Turkish Translation

r3dmc:


Scarsz:
Do you not want to do the resources as well, @r3dmc?

PR #594, opened 2019-12-17
Fix nickname sync

underscore11code:


underscore11code:
Scarsz wants it done differently

PR #595, opened 2019-12-18
Simplifying placeholders

underscore11code:
This consolidates all common placeholder replace methods into 1 file, to allow for easy addition of new placeholders, less repeated code, and more placeholders available for the end user.

All placeholders now follow a style of `%CATEGORY_FIELD%`. Each category is a different entity e.g. Player, Server, Guild, User, with different fields for each.

End goal is that each field that supports placeholders in config files can simply say which categories it supports, as well as any special placeholders specific to that field (e.g. %level% for console channel)

As of original Pull Request, comments in config files have not been edited yet

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -47,7 +47,6 @@ import github.scarsz.discordsrv.objects.threads.ConsoleMessageQueueWorker;
 import github.scarsz.discordsrv.objects.threads.PresenceUpdater;
 import github.scarsz.discordsrv.objects.threads.ServerWatchdog;
 import github.scarsz.discordsrv.util.*;
-import javafx.print.PageLayout;
 import lombok.Getter;
 import me.vankka.reserializer.discord.DiscordSerializer;
 import me.vankka.reserializer.minecraft.MinecraftSerializer;
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -47,6 +47,7 @@ import github.scarsz.discordsrv.objects.threads.ConsoleMessageQueueWorker;
 import github.scarsz.discordsrv.objects.threads.PresenceUpdater;
 import github.scarsz.discordsrv.objects.threads.ServerWatchdog;
 import github.scarsz.discordsrv.util.*;
+import javafx.print.PageLayout;
 import lombok.Getter;
 import me.vankka.reserializer.discord.DiscordSerializer;
 import me.vankka.reserializer.minecraft.MinecraftSerializer;
@@ -902,6 +903,8 @@ public class DiscordSRV extends JavaPlugin implements Listener {
                 .replace("%worldalias%", DiscordUtil.strip(MultiverseCoreHook.getWorldAlias(player.getWorld().getName())))
         ;
         discordMessage = PlaceholderUtil.applyAll(player, null, discordMessage, true);
+        discordMessage = PlaceholderUtil.applyLegacy("%primarygroup", userPrimaryGroup, discordMessage, "%minecraft_group");
+        discordMessage = PlaceholderUtil.applyLegacy("%username%", username, discordMessage, "%minecraft_playername%");
 
         String displayName = DiscordUtil.strip(player.getDisplayName());
         if (!reserializer) displayName = DiscordUtil.escapeMarkdown(displayName);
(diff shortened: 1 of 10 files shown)


underscore11code:
Forgot to hit "Draft Pull Request", until I change config files, treat this like it isn't ready to pull.
Feel free to yell at me about code style or give other feedback in the meantime

Scarsz:
This is a very breaking change

underscore11code:
Yeah fair enough. However, would be quite easily expandable in the future.

*And people would stop complaining about not being able to use full Name#discrim in message format.*

src/main/java/github/scarsz/discordsrv/util/PlaceholderUtil.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+// A bunch of general placeholders.
+public class PlaceholderUtil {
+    // Tries to apply placeholderAPI placeholders
+    public static String applyPlaceholderApi(Player player, String text) {
+        if (PluginUtil.pluginHookIsEnabled("placeholderapi")) return me.clip.placeholderapi.PlaceholderAPI.setPlaceholders(player, text);
```
Vankka: While you're at it, PAPI translates colors; thus we should translate them back (& -> §)

underscore11code: PlaceholderAPI is used for discord as well however, so we'd get a lot of unnecessary colour codes in discord

Vankka: what i'm saying is for Discord messages; role mentions break because of PAPI because it's converting <@&id> to <@§id>

underscore11code:
Scarsz, what if I was to add a backwards compatibility method for the old placeholders, but have it send warns if it detects any that they will not be supported in a coming update, and should be swapped out for the new ones.

Would that work?

Vankka:
I don't think removing them is necessary at all, Iris (rewrite) is planned for the upcoming summer

src/main/java/github/scarsz/discordsrv/util/PlaceholderUtil.java
underscore11code: Oh I see. How could we work around that though? Only thing I can think of is replacing the `&` with some other unicode character that people are unlikely to use, then replacing it back after PAPI does it's thing.

Or maybe a placeholder like %role_ROLEID% that subs it out for a proper role mention after all the other placeholders are done...that makes more sense.

src/main/java/github/scarsz/discordsrv/util/PlaceholderUtil.java
```diff
@@ -121,6 +121,14 @@ public static String applyAll(Player player, User user, String text) {
         return applyAll(player, user, text, false);
     }
 
+    public static String applyUnsupported(String legacyPlaceholder, String data, String text, String newPlaceholder) {
+        if (text.contains(legacyPlaceholder)) {
+            DiscordSRV.info("Found legacy placeholder "+legacyPlaceholder+" in "+text+". This should be replaced with "+newPlaceholder);
```
Vankka: Spaces before and after pluses, please

underscore11code: fixed in eac9132

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -902,6 +903,8 @@ public void processChatMessage(Player player, String message, String channel, bo
                 .replace("%worldalias%", DiscordUtil.strip(MultiverseCoreHook.getWorldAlias(player.getWorld().getName())))
         ;
         discordMessage = PlaceholderUtil.applyAll(player, null, discordMessage, true);
+        discordMessage = PlaceholderUtil.applyLegacy("%primarygroup", userPrimaryGroup, discordMessage, "%minecraft_group");
```
Vankka: missing % at the end

src/main/java/github/scarsz/discordsrv/objects/threads/ChannelTopicUpdater.java
```diff
@@ -59,4 +60,18 @@ public void run() {
             }
         }
     }
+
+    private String applyLegacyPlaceholders(String text) {
+        String[] legacyPlaceholders = {"playercount","playermax","totalplayers","uptimemins","uptimehours","uptimedays",
+                "motd","serverversion","freememory","usedmemory","totalmemory","maxmemory","freememorygb","usedmemorygb","totalmemorygb","maxmemorygb","tps"};
+        for (String legacyPlaceholder : legacyPlaceholders) {
+            if (text.contains(legacyPlaceholder))
+                DiscordSRV.info("Found legacy placeholder " + legacyPlaceholder + " in " + text + ". This should be replaced with %server_" + legacyPlaceholder + "%");
```
Vankka: - This is bad as a info message (could be triggered frequently depending on activity of a server), 
- This check doesn't have % signs at the end or beginning, 
- That array could be made a variable in the class instead of in the method to save it from initiating it every time. 

We should skip 'deprecating' placeholders at this stage of v1, they should keep working like they do now

underscore11code:
I'm going to close this, since this is more of a rewrite. Maybe I'll stick this in Iris later once there is more of a codebase.

PR #601, opened 2019-12-24
Add Invite command

underscore11code:
An easy command to generate a bot invite, eliminating the need for people to use https://scarsz.me/authorize.

Would change install workflow to: 

1. Paste jar into plugins folder
2. Start and stop server to generate files
3. Fill in config completely
4. Start server, `/discordsrv:discord invite`, `/discordsrv:discord reload`

This workflow has been tested and does work.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/commands/CommandInvite.java
+++ b/src/main/java/github/scarsz/discordsrv/commands/CommandInvite.java
@@ -11,7 +11,11 @@ public class CommandInvite {
             permission = "discordsrv.invite"
     )
     public static void execute(CommandSender sender, String[] args) {
-        String invite = DiscordUtil.getJda().getInviteUrl(Permission.ADMINISTRATOR);
-        sender.sendMessage(ChatColor.DARK_AQUA + "You may invite the bot with this link:\n" + ChatColor.AQUA + invite);
+        try {
+            String invite = DiscordUtil.getJda().getInviteUrl(Permission.ADMINISTRATOR);
+            sender.sendMessage(ChatColor.DARK_AQUA + "You may invite the bot with this link:\n" + ChatColor.AQUA + invite);
+        } catch (NullPointerException e) {
+            sender.sendMessage(ChatColor.RED + "Could not generate an invite. Is your token valid?");
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/resources/plugin.yml
```diff
@@ -98,4 +98,7 @@ permissions:
     default: true
   discordsrv.linked.others:
     description: whether or not the player is able to check what Discord account other Minecraft accounts are linked to
+    default: op
+  discordsrv.invite:
+    description: whether or not the player is able to generate a bot invite link
```
Vankka: -> gives an invite url for the bot of this DiscordSRV instance (or simmilar)

underscore11code:
Fixed a couple omitted things from plugin.yml, now properly checks for null JDA instead of catching NPEs blindly

Vankka:
This isn't perfect, this requires the bot token to be correct; quite a lot of people come for support with "bot is offline"

src/main/java/github/scarsz/discordsrv/commands/CommandInvite.java
```diff
@@ -0,0 +1,21 @@
...
+import org.bukkit.command.CommandSender;
+
+public class CommandInvite {
+    @Command(commandNames = { "invite" },
+            helpMessage = "Generates an invite code for the bot",
+            permission = "discordsrv.invite"
+    )
+    public static void execute(CommandSender sender, String[] args) {
+        if (DiscordUtil.getJda() != null) {
+            String invite = DiscordUtil.getJda().getInviteUrl(Permission.ADMINISTRATOR);
```
Andre601: This should really be avoided imo.
Someone just needs to get the token to mess up the entire Discord server.

Just give back an invite with no permissions set (Similar to https://scarsz.me/authorize)
```suggestion
            String invite = DiscordUtil.getJda().getInviteUrl();
```

Vankka: This is a whole other conversation, the Installation guide already instructs users to create a role with Administrator permissions; or individual permissions for _advanced users_.

Andre601: Then the guide should be updated.
Like I said should we stop with this habit of giving out admins to bots.
If people won't spend time to setup perms should they also not run srvers.

pkrok01: The Installation guide says (in a condensed format) `Add the Administrator permission to the bot. If you're experienced, here's a list of permissions the bot needs`. It's not our job to dictate what others do with their discord servers. I do support your commit suggestion, though.

Scarsz:
Not gonna merge this because eventually DiscordSRV is going to have a configuration GUI and this could just be handled through that instead

PR #625, opened 2020-02-05
 Implement cut phrases on webhooks

samsamson33:
Pretty much just copy/pasting the code from `DiscordUtil.java` to `WebhookUtil.java`. This is an issue since, when webhooks are enabled, there's no way to stop only `@everyone` and `@here` mentions while allowing other mentions.


Scarsz:
Already been implemented in develop

PR #643, opened 2020-02-23
Log rotation when using DiscordConsoleChannelUsageLog

### LINKED ISSUE #627: Enhancement: Log rotation when using DiscordConsoleChannelUsageLog
Allow for per-day log names for inbuild log rotation.

Preferably with inbuilt gzipping of the previous day's log and an option to remove logs past a certain date, but appreciate these two requests are optional and can be solved easily by the use of other plugins.

metalshark:
Added PlaceholderAPI support to DiscordConsoleChannelUsageLog values.
Fixes #627

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -339,18 +339,18 @@ public class DiscordChatListener extends ListenerAdapter {
 
         // log command to console log file, if this fails the command is not executed for safety reasons unless this is turned off
         File logFile = DiscordSRV.getPlugin().getLogFile();
-        try {
-            if (logFile != null) {
+        if (logFile != null) {
+            try {
                 FileUtils.writeStringToFile(
-                        logFile,
-                        "[" + TimeUtil.timeStamp() + " | ID " + event.getAuthor().getId() + "] " + event.getAuthor().getName() + ": " + event.getMessage().getContentRaw() + System.lineSeparator(),
-                        StandardCharsets.UTF_8,
-                        true
+                    logFile,
+                    "[" + TimeUtil.timeStamp() + " | ID " + event.getAuthor().getId() + "] " + event.getAuthor().getName() + ": " + event.getMessage().getContentRaw() + System.lineSeparator(),
+                    StandardCharsets.UTF_8,
+                    true
                 );
+            } catch (IOException e) {
+                DiscordSRV.error(LangUtil.InternalMessage.ERROR_LOGGING_CONSOLE_ACTION + " " + logFile.getAbsolutePath() + ": " + e.getMessage());
+                if (DiscordSRV.config().getBoolean("CancelConsoleCommandIfLoggingFailed")) return true;
             }
-        } catch (IOException e) {
-            DiscordSRV.error(LangUtil.InternalMessage.ERROR_LOGGING_CONSOLE_ACTION + " " + DiscordSRV.config().getString("DiscordConsoleChannelUsageLog") + ": " + e.getMessage());
-            if (DiscordSRV.config().getBoolean("CancelConsoleCommandIfLoggingFailed")) return true;
         }
 
         // at this point, the user has permission to run commands at all and is able to run the requested command, so do it
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -40,8 +40,6 @@ import org.bukkit.entity.Player;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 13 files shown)


metalshark:
@Vankka more like that?

src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
```diff
@@ -82,7 +84,14 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
 
         // log command to console log file, if this fails the command is not executed for safety reasons unless this is turned off
         try {
-            String fileName = DiscordSRV.config().getString("DiscordConsoleChannelUsageLog");
+            DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+            LocalDateTime now = LocalDateTime.now();
+            File dataFolder = DiscordSRV.getPlugin().getDataFolder();
+            if (!dataFolder.exists()) dataFolder.mkdir();
+            File logFolder = new File(dataFolder.getAbsolutePath() + File.separator + "logs");
```
Vankka:
```suggestion
            File logFolder = new File(dataFolder, "logs");
```

src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
```diff
@@ -82,7 +84,14 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
+            if (!logFolder.exists()) logFolder.mkdir();
+            String fileName = logFolder + File.separator + "console-log-" + dtf.format(now) + ".log";
+
             if (StringUtils.isNotBlank(fileName)) {
```
Vankka: This check is very invalid now

src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
```diff
@@ -82,7 +84,14 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
         // log command to console log file, if this fails the command is not executed for safety reasons unless this is turned off
         try {
-            String fileName = DiscordSRV.config().getString("DiscordConsoleChannelUsageLog");
+            DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+            LocalDateTime now = LocalDateTime.now();
+            File dataFolder = DiscordSRV.getPlugin().getDataFolder();
+            if (!dataFolder.exists()) dataFolder.mkdir();
+            File logFolder = new File(dataFolder.getAbsolutePath() + File.separator + "logs");
+            if (!logFolder.exists()) logFolder.mkdir();
+            String fileName = logFolder + File.separator + "console-log-" + dtf.format(now) + ".log";
```
Vankka: Change this to `File file = new File(logFolder, "console-log-" + dtf.format(now) + ".log");`

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -338,15 +340,23 @@ private boolean processConsoleCommand(GuildMessageReceivedEvent event, String me
         }
 
         // log command to console log file, if this fails the command is not executed for safety reasons unless this is turned off
+        DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+        LocalDateTime now = LocalDateTime.now();
+        File dataFolder = DiscordSRV.getPlugin().getDataFolder();
+        if (!dataFolder.exists()) dataFolder.mkdir();
+        File logFolder = new File(dataFolder.getAbsolutePath() + File.separator + "logs");
+        if (!logFolder.exists()) logFolder.mkdir();
+        String fileName = logFolder + File.separator + "console-log-" + dtf.format(now) + ".log";
```
Vankka: Same changes here as the other file.
Also copy pasting code like this usually isn't a good idea; central method would be nice eg. `File getLogFile()`

src/main/resources/config/en.yml
```diff
@@ -130,7 +129,6 @@ DiscordChatChannelRolesSelection: ["Don't show me!", "Misc role"]
 # DiscordConsoleChannelLevels: levels to send to console channel via appender
 #
 DiscordConsoleChannelLogRefreshRateInSeconds: 5
-DiscordConsoleChannelUsageLog: "DiscordConsole.log"
```
Vankka: Removal might not've been ideal, this could've allowed changing between 3 settings;
- Disabled (no logging)
- One file (just like the old system, just one file for all the logging) [Though, the folder would be in the way in this case]
- Multiple files (eg. with a placeholder like `%date%`)

metalshark: That's why I was thinking PlaceholderAPI before. Allows you to per player log files, etc

src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
metalshark: With the try/catch, later on, should I ignore the folder exists checks?

Vankka: what

src/main/resources/config/en.yml
Vankka: Well that's a possibility, but rolling with date shouldn't require PAPI on it's own

src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
metalshark: Sorry am not too familiar with Java, I don't understand what you mean by "This check is very invalid now".

Vankka: `fileName` is never blank, so that will always be `true`

PR #644, opened 2020-02-23
Separated the Discord invite link into it's own config setting. 

### LINKED ISSUE #629: Define invite link in config.yml then refer to it as a placeholder
It would be nice to define the invite link once in the main config.yml then refer to it in:

linking.yml
Require linked account to play > Not linked message

and

messages.yml
DiscordCommandFormat

as a placeholder. Adding a note to remind admins to create a "never expires" invite might also be friendly move.

metalshark:
Separated the Discord invite link into its own config setting.
Fixes #629


src/main/resources/config/en.yml
```diff
@@ -16,6 +16,8 @@ Channels: {"global": "000000000000000000"}
 # Console channel numerical ID (NOT NAME), leave blank to disable the console channel all together
 DiscordConsoleChannelId: "000000000000000000"
 
+DiscordInviteLink: ""
```
Vankka: Move to messages.yml below/near `DiscordCommandFormat`, add a comment to the option, default value of `discord.gg/changethisinthemessages.yml` (or something along those lines)

src/main/resources/linking/en.yml
```diff
@@ -21,7 +21,7 @@ Require linked account to play:
   # Message to kick players with telling them to link their accounts
   # Use {BOT} as a placeholder for the bot's name
   # Use {CODE} as a placeholder for the code people need to DM the bot
-  Not linked message: "&7You must link your &9Discord &7account to play.\n\n&7Send a DM to &b{BOT}&7 in the Discord server containing just &b{CODE}&7 to link your account.\n\n&7Discord Invite » &bhttps://discord.gg/changethisinlinking.yml"
+  Not linked message: "&7You must link your &9Discord &7account to play.\n\n&7Send a DM to &b{BOT}&7 in the Discord server containing just &b{CODE}&7 to link your account.\n\n&7Discord Invite » &b{INVITE}"
```
Vankka: Add explanation of {INVITE} in the comment & where it is (config.yml)

src/main/resources/messages/en.yml
```diff
@@ -186,7 +186,7 @@ ChannelTopicUpdaterConsoleChannelTopicAtServerShutdownFormat: "Server is offline
 # Discord command message
 # This is the message sent to players when they run "/discord". It's recommended to leave command syntax as a part of this
 #
-DiscordCommandFormat: "&bJoin us on Discord at http://discord.gg/changethisinmessages.yml. For command help, do \"/discord ?\""
+DiscordCommandFormat: "&bJoin us on Discord at {INVITE}. For command help, do \"/discord ?\""
```
Vankka: Add explanation of {INVITE} in the comment & that it's a config option in the same file (in some logical, user-friendly way)

src/main/resources/config/en.yml
metalshark: No problem with the changing of the default value, but moving it to the messages.yml file removes the benefit of this patch, e.g. to separate config from messages.

Vankka: The benafit is that the option can be changed for 2 options at once, that is still there. Putting it in a entirely different place doesn't make sense to me, and putting it near the Discord command format makes sense; most people configure that, unlike require linked 4 gameplay that is used by ~38 servers out of 2000+ on record (https://bstats.org/plugin/bukkit/DiscordSRV/387)

metalshark: Being able to exchange messages.yml files with others, having tailored the language to tailor the style, without it containing server related configuration and settings is the goal of this. De-duplication is an added bonus.

Vankka: Ok, that's a valid point

src/main/resources/linking/en.yml
```diff
@@ -21,6 +21,7 @@ Require linked account to play:
   # Message to kick players with telling them to link their accounts
   # Use {BOT} as a placeholder for the bot's name
   # Use {CODE} as a placeholder for the code people need to DM the bot
+  # Use {INVITE} as a placeholder for the invite link people need to join the Discord server
```
Vankka: This should include where the option is, or even what the option is called

src/main/resources/messages/en.yml
```diff
@@ -185,6 +185,7 @@ ChannelTopicUpdaterConsoleChannelTopicAtServerShutdownFormat: "Server is offline
 
 # Discord command message
 # This is the message sent to players when they run "/discord". It's recommended to leave command syntax as a part of this
+# Use {INVITE} as a placeholder for the invite link people need to join the Discord server
```
Vankka: This should include where the option is, or even what the option is called

src/main/resources/config/en.yml
```diff
@@ -16,6 +16,8 @@ Channels: {"global": "000000000000000000"}
 # Console channel numerical ID (NOT NAME), leave blank to disable the console channel all together
 DiscordConsoleChannelId: "000000000000000000"
 
+DiscordInviteLink: "discord.gg/changethisintheconfig.yml"
```
Vankka: Comment of where the option is used

PR #645, opened 2020-02-23
Added new DiscordServerAdmin option to help keep configuration separate from messaging

### LINKED ISSUE #630: Define server admin in config.yml and refer to it as a placeholder
Being able to define the primary server admin in config.yml then refer to it in:

messages.yml
ServerWatchdogMessage

with it having a default of `%guildowner%` may cut down on future support requests. Admins can still bespoke the ServerWatchdogMessage or add more than one server admin (using the <@123> format).

metalshark:
Added new DiscordServerAdmin option to help keep configuration separate from messaging.
Fixes #630


Vankka:
I don't see thing being necessary, as it is only used in one place; now it is just split apart into two options instead of just one

metalshark:
Where a Minecraft server is retrofitted onto an existing Discord it makes it easier to separate config from messages. If this gets merged will ask for it to be added to the placeholders as well to make it more modular for sharing configuration between servers.

metalshark:
Remember also this is a repeat question in the support Discord.

Vankka:
as long as the watchdog is enabled by default, it will keep being a repeat question, this won't effect that

metalshark:
What I mean is the question of how you configure who gets notified. Normally when you're done with a plugin's config.yml you don't need to go into the language files to set configuration.

Scarsz:
I'm not particularly a fan of this or the reason provided for the change. Having an entire configuration option that is only utilized in one message as opposed to just being in the message is unnecessary. While I can see the point of view that the config and messages files can be separated, realistically, it's not necessary for such a small use.

PR #649, opened 2020-02-24
Debug config options cleanup, and splitting of options

Vankka:
Goals:
- [x] Make it clear that RespectChatPlugins shouldn't be used in production
- [x] Cleanup the debug options section in the config
    + [x] For other languages
- [x] Separate debug options (as per Scarsz' todo in https://github.com/DiscordSRV/DiscordSRV/pull/624)
- [x] translate new debug options in other translation files  
- [x] Test This PR in case something broke in all the rebases and merges
- [x] ~~Create a PR for~~ Update Audrey to accommodate for the new debug option values
- [ ] Prepare canned responses / support bot questions based on new debug option values
- [x] Talk about the `MinecraftDiscordAccount...` options in below comment


src/main/java/github/scarsz/discordsrv/util/LangUtil.java
```diff
@@ -196,6 +196,97 @@
...
+                    "RespectChatPlugins가 비활성화되었습니다.이 옵션은 테스트 목적으로 만 사용됩니다.\n" +
+                    "프로덕션 서버에서는 절대 비활성화하지 않아야합니다.\n" +
+                    "이 옵션을 비활성화하면 취소 된 메시지가 불일치로 전달됩니다.\n" +
+                    "/command가없는 개인 메시지 또는 직원 채팅 메시지를 포함하지만 이에 국한되지는 않습니다.\n" +
+                    "\n" +
+                    "\n");
+            put(Language.NL,
+                    "\n" +
+                    "\n" +
+                    "RespectChatPlugins is uitgeschakeld, deze optie is ALLEEN voor TESTEN VAN DOELEINDEN\n" +
```
LemmoTresto, line 248: Make this `RespectChatPlugins is uitgeschakeld, deze optie is ALLEEN voor TESTEN VAN DOELEINDEN\n` into `RespectChatPlugins is uitgeschakeld, deze optie is ALLEEN voor TEST DOELEINDEN\n`

Vankka: Thanks, pro tip there's the "Insert a suggestion" button, the + and - on top of each other

src/main/resources/linking/en.yml
```diff
@@ -1,3 +1,25 @@
...
+#                         example: Notch
+#
+# MinecraftDiscordAccountLinkedRoleNameToAddUserTo: the name of a discord role to add a discord user to when they link their account
+# MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: allows sending a new code to the bot to unlink and relink with the new code
+#
+MinecraftDiscordAccountLinkedConsoleCommands: ["", "", ""]
+MinecraftDiscordAccountUnlinkedConsoleCommands: ["", "", ""]
+MinecraftDiscordAccountLinkedRoleNameToAddUserTo: "Linked"
+MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: false
+
```
granny, line 22: Should we be moving `MinecraftDiscordAccount...` options into `linking.yml`? I don't remember who recommended (probably me, since I committed it) but I'm not sure it's a good idea anymore (unless at some point option migration between config files was implemented?) Would probably be best to just move them back into `config.yml`

granny (changes requested):
[ ] Test This PR in case something broke in all the rebases and merges
[ ] Create a PR for Audrey that's updated to accommodate for the new debug option values
[ ] prepare canned responses / support bot questions based on new debug option values
[ ] Talk about the `MinecraftDiscordAccount...` options in below comment

~~(Files changed: nice)~~

Vankka:
>  Test This PR in case something broke in all the rebases and merges
I checked all the changed files (besides configs), and confirmed that there were no changes apart from what was intended. I have also verified that debug messages were categories where there was a fitting debug category in my most recent merge. So as far as I'm concerned this is done, though it's probably a good idea to have someone else check it over too.

>  Create a PR for Audrey that's updated to accommodate for the new debug option values
Maybe the the list of debug categories should also be listed in debug scan results, or have a warning (or even info) if `ALL` isn't used specifically. Currently the debug is not enabled error will only be active if there are no debug categories specified.

src/main/resources/linking/en.yml
Vankka: Config migration is copies lines from the same file as of now. For existing users moving it might confuse them if they can't find the option anymore, for new users it might help with finding those options. 

It's probably not a good idea to reset those options for people, so I'd suggest moving those back to the config.

Vankka: https://github.com/DiscordSRV/DiscordSRV/commit/544f53ad4814a2e1ab6c9cb563a7f7af5165a891

Dinty1 (approved):
Rebased :)

PR #651, opened 2020-02-25
Added achievement message events

### LINKED ISSUE #650: Add AchievementMessageEvents to the API
Being able to change the achievement text using a lookup table in a 3rd party plugin is desirable and this will require an API to hook into.

metalshark:
Added achievement message events.
Fixes #650

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
@@ -59,11 +59,7 @@ public class PlayerAchievementsListener implements Listener {
         String message = LangUtil.Message.PLAYER_ACHIEVEMENT.toString();
         Player player = event.getPlayer();
 
-        AchievementMessagePreProcessEvent preEvent = DiscordSRV.api.callEvent(new AchievementMessagePreProcessEvent(channelName, message, player, achievementName, false));
-        if (preEvent.isCancelled()) {
-            DiscordSRV.debug("AchievementMessagePreProcessEvent was cancelled, message send aborted");
-            return;
-        }
+        AchievementMessagePreProcessEvent preEvent = DiscordSRV.api.callEvent(new AchievementMessagePreProcessEvent(channelName, message, player, achievementName));
         // Update from event in case any listeners modified parameters
         achievementName = preEvent.getAchievementName();
         channelName = preEvent.getChannel();
@@ -78,7 +74,7 @@ public class PlayerAchievementsListener implements Listener {
                 .replace("%achievement%", achievementName);
         message = PlaceholderUtil.replacePlaceholdersToDiscord(message, player);
 
-        AchievementMessagePostProcessEvent postEvent = DiscordSRV.api.callEvent(new AchievementMessagePostProcessEvent(channelName, message, player, achievementName, false));
+        AchievementMessagePostProcessEvent postEvent = DiscordSRV.api.callEvent(new AchievementMessagePostProcessEvent(channelName, message, player, achievementName, preEvent.isCancelled()));
         if (postEvent.isCancelled()) {
             DiscordSRV.debug("AchievementMessagePostProcessEvent was cancelled, message send aborted");
             return;
--- a/src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
@@ -53,8 +53,9 @@ public class PlayerAchievementsListener implements Listener {
         if (PlayerUtil.isVanished(event.getPlayer())) return;
 
         // turn "ACHIEVEMENT_NAME" into "Achievement Name"
+        TextChannel channel = DiscordSRV.getPlugin().getMainTextChannel();
+        String channelName = channel.getName();
         String achievementName = PrettyUtil.beautify(event.getAchievement());
-        String channelName = DiscordSRV.getPlugin().getMainTextChannel().getName();
         String message = LangUtil.Message.PLAYER_ACHIEVEMENT.toString();
         Player player = event.getPlayer();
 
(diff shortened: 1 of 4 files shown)


src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
```diff
@@ -50,17 +54,41 @@ public void onPlayerAchievementAwarded(PlayerAchievementAwardedEvent event) {
 
         // turn "ACHIEVEMENT_NAME" into "Achievement Name"
         String achievementName = PrettyUtil.beautify(event.getAchievement());
+        String channelName = DiscordSRV.getPlugin().getMainTextChannel().getName();
```
Vankka: This would return the name of the channel on Discord, but you later retrieve the in-game channel using this variable.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
```diff
@@ -59,21 +64,47 @@ public void onPlayerAdvancementDone(PlayerAdvancementDoneEvent event) {
             return;
         }
 
+        String channelName = DiscordSRV.getPlugin().getMainTextChannel().getName();
```
Vankka: This would return the name of the channel on Discord, but you later retrieve the in-game channel using this variable.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
```diff
@@ -59,21 +64,46 @@ public void onPlayerAdvancementDone(PlayerAdvancementDoneEvent event) {
             return;
         }
 
+        TextChannel channel = DiscordSRV.getPlugin().getMainTextChannel();
+        String channelName = channel.getName();
```
Vankka: This should be the in-game channel, not Discord.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
```diff
@@ -49,18 +53,41 @@ public void onPlayerAchievementAwarded(PlayerAchievementAwardedEvent event) {
         if (PlayerUtil.isVanished(event.getPlayer())) return;
 
         // turn "ACHIEVEMENT_NAME" into "Achievement Name"
+        TextChannel channel = DiscordSRV.getPlugin().getMainTextChannel();
+        String channelName = channel.getName();
```
Vankka: This should be the in-game channel, not Discord.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
```diff
@@ -59,21 +64,46 @@ public void onPlayerAdvancementDone(PlayerAdvancementDoneEvent event) {
...
+
         // turn "story/advancement_name" into "Advancement Name"
-        String rawAdvancementName = event.getAdvancement().getKey().getKey();
+        String rawAdvancementName = advancement.getKey().getKey();
         String advancementName = Arrays.stream(rawAdvancementName.substring(rawAdvancementName.lastIndexOf("/") + 1).toLowerCase().split("_"))
                 .map(s -> s.substring(0, 1).toUpperCase() + s.substring(1))
                 .collect(Collectors.joining(" "));
 
-        String discordMessage = LangUtil.Message.PLAYER_ACHIEVEMENT.toString()
+        AchievementMessagePreProcessEvent preEvent = DiscordSRV.api.callEvent(new AchievementMessagePreProcessEvent(channelName, message, player, advancementName));
```
Vankka: Cancellation should be checked

src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
```diff
@@ -49,18 +53,41 @@ public void onPlayerAchievementAwarded(PlayerAchievementAwardedEvent event) {
...
         String achievementName = PrettyUtil.beautify(event.getAchievement());
+        String message = LangUtil.Message.PLAYER_ACHIEVEMENT.toString();
+        Player player = event.getPlayer();
 
-        String discordMessage = LangUtil.Message.PLAYER_ACHIEVEMENT.toString()
+        AchievementMessagePreProcessEvent preEvent = DiscordSRV.api.callEvent(new AchievementMessagePreProcessEvent(channelName, message, player, achievementName));
```
Vankka: Cancellation should be checked

metalshark: Changed this to be consistent with your other uses (consistent API). So the results of it being cancelled are passed into the post-processing event, then the checking of the message being cancelled is done after the post-processing event.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
metalshark: Changed this to be consistent with your other uses (consistent API). So the results of it being cancelled are passed into the post-processing event, then the checking of the message being cancelled is done after the post-processing event.

src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
metalshark: The previous code used `DiscordSRV.getPlugin().getMainTextChannel()` am I not using the same thing here? If not what method would you prefer I got the channel using?

Vankka: getMainChatChannel()

src/main/java/github/scarsz/discordsrv/listeners/PlayerAchievementsListener.java
Vankka: Doesn't appear to be the case, checked all events with Pre & Post process scheme, they all would return if preprocess was cancelled

metalshark: https://github.com/DiscordSRV/DiscordSRV/blob/develop/src/main/java/github/scarsz/discordsrv/DiscordSRV.java#L963

metalshark: No, you're right.

PR #658, opened 2020-03-08
Added way to disable group synchronization, to reduce extra processing

### LINKED ISSUE #657: Provide an option to disable group synchronization code paths entirely
The `resync` function takes a long amount of time to run on our main server, and we use an alternative method of synchronisation.

metalshark:
Added GroupRoleSynchronizationEnabled option to prevent unnecessary resync work when unused.
Fixes #657

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -1077,4 +1077,26 @@ public class DiscordSRV extends JavaPlugin implements Listener {
                 !config().getBooleanElse("UpdateCheckDisabled", false);
     }
 
+    public static final String DEFAULT_ROLE_ID = "000000000000000000";
+    public static final Map<String, String> DEFAULT_GROUP_AND_ROLES_TO_SYNC = new HashMap<String, String>() {{
+        put("trusted", DEFAULT_ROLE_ID);
+        put("vip", DEFAULT_ROLE_ID);
+    }};
+
+    /**
+     * @return Whether or not DiscordSRV group role synchronization has been enabled in the configuration.
+     */
+    public static boolean isGroupRoleSynchronizationEnabled() {
+        final Map<String, String> groupsAndRolesToSync = DiscordSRV.config().getMap("GroupRoleSynchronizationGroupsAndRolesToSync");
+        if (groupsAndRolesToSync.isEmpty()) return false;
+        if (groupsAndRolesToSync.equals(DEFAULT_GROUP_AND_ROLES_TO_SYNC)) return false;
+        for(Map.Entry<String, String> entry : groupsAndRolesToSync.entrySet()) {
+            final String group = entry.getKey();
+            final String roleId = entry.getValue();
+            if (group.isEmpty()) return false;
+            if (roleId.isEmpty() || roleId.equals(DEFAULT_ROLE_ID)) return false;
+        }
+        return true;
+    }
+
 }
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -1087,9 +1087,9 @@ public class DiscordSRV extends JavaPlugin implements Listener {
         if (groupsAndRolesToSync.isEmpty()) return false;
         for(Map.Entry<String, String> entry : groupsAndRolesToSync.entrySet()) {
             final String group = entry.getKey();
-            final String roleId = entry.getValue();
             if (!group.isEmpty()) {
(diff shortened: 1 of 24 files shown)


src/main/java/github/scarsz/discordsrv/commands/CommandResync.java
```diff
@@ -14,14 +14,16 @@
...
+        if (DiscordSRV.config().getBoolean("GroupRoleSynchronizationEnabled")) {
+            Bukkit.getScheduler().runTaskAsynchronously(DiscordSRV.getPlugin(), () -> {
+                sender.sendMessage(ChatColor.AQUA + "Full group synchronization triggered.");
+                long time = System.currentTimeMillis();
+                DiscordSRV.getPlugin().getGroupSynchronizationManager().resyncEveryone();
+                time = System.currentTimeMillis() - time;
+                int seconds = Math.toIntExact(TimeUnit.MILLISECONDS.toSeconds(time));
+                sender.sendMessage(ChatColor.AQUA + "Full group synchronization finished, taking " + seconds + " seconds.");
+            });
+        }
```
Vankka: Should provide a reponse if disabled

src/main/java/github/scarsz/discordsrv/commands/CommandResync.java
```diff
@@ -14,6 +14,10 @@
             permission = "discordsrv.resync"
     )
     public static void execute(CommandSender sender, String[] args) {
+        if (!DiscordSRV.config().getBoolean("GroupRoleSynchronizationEnabled")) {
+            sender.sendMessage(ChatColor.RED + "Group synchonization is disabled. Please set GroupRoleSynchronizationEnabled to true in synchronization.yml to use this feature.");
```
Vankka: LangUtil.InternalMessage

Vankka:
Commands may also make groups to resync

Scarsz:
This should just check if there’s no groups to sync in the config list, a dedicated config option to disable isn’t necessary

src/main/java/github/scarsz/discordsrv/commands/CommandResync.java
```diff
@@ -14,6 +15,10 @@
             permission = "discordsrv.resync"
     )
     public static void execute(CommandSender sender, String[] args) {
+        if (DiscordSRV.config().getMap("GroupRoleSynchronizationGroupsAndRolesToSync").isEmpty()) {
```
Vankka: Would be better to also check that the ids aren't/are valid incase the option is on default values

metalshark: Included a check on default and a check for roleId. Feel free to pick which you'd prefer (as both seems unnecessary).

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -1077,4 +1077,26 @@ public static boolean isUpdateCheckDisabled() {
...
+     */
+    public static boolean isGroupRoleSynchronizationEnabled() {
+        final Map<String, String> groupsAndRolesToSync = DiscordSRV.config().getMap("GroupRoleSynchronizationGroupsAndRolesToSync");
+        if (groupsAndRolesToSync.isEmpty()) return false;
+        if (groupsAndRolesToSync.equals(DEFAULT_GROUP_AND_ROLES_TO_SYNC)) return false;
+        for(Map.Entry<String, String> entry : groupsAndRolesToSync.entrySet()) {
+            final String group = entry.getKey();
+            final String roleId = entry.getValue();
+            if (group.isEmpty()) return false;
+            if (roleId.isEmpty() || roleId.equals(DEFAULT_ROLE_ID)) return false;
```
Vankka: This returns false even when there is a valid id and one or more default values

metalshark: We have choices:
1. only look for default configuration and/or completely blank
2. look for any role being 18 x 0's
3. look for at least one role not being 18 x 0's
am guessing you'd prefer option 3?

Vankka: number 3 would be ideal here, yes

metalshark: How do you feel about the constant `DEFAULT_ROLE_ID` being declared? Can easily get rid of it and just put the string in instead.

Vankka: doesn't really make a difference, though using a constant is inconsistent with other checks in DiscordSRV

metalshark:
Thinking about this again, I should probably cache the result of the `isGroupRoleSynchronizationEnable()` check and refresh it when the config is reloaded.

Vankka:
Would be inconsistent with other implementations, and it's a fairly minor speed bump

PR #672, opened 2020-03-31
Made Voice module skip tick if the previous one hasn't finished, making it thread safe

metalshark:
Changed dirtyPlayers to ConcurrentHashSet, synchronized whole tick method and cloned dirtyPlayers before loop. Fixes issue 669.


Vankka:
What benafit does synchronizing the tick() method bring?

metalshark:
Prevents two ticks being run concurrently (kinda defeats there being a
tick). Having logic to cancel (skip) a tick being run if the previous one
is still being processed actually makes more sense.

Vankka:
I'd think synchronization would just make the ticks back up if they took too long, so if tick one takes too long: tick 2 would have to wait, and will sort of be added to a queue. So if all the ticks take too long, the "queue" will just get filled

Vankka:
Cancelling the previous tick might not be ideal either. Would rather skip a whole tick if the previous tick takes too long, and possibly have a maximum time a tick can take in case that's a concern

metalshark:
Yup skipping a tick is the way to go. Will do some testing with that way to handle it.

Vankka:
Mostly a preference thing but, a Semaphore with a size of 1 works the same way a ReentrantLock would. Instead of a Closable type `try { ... } finally { lock.close() }` could be used. 

I'm fine with this currently, unless other have something to comment we can get this merged

metalshark:
Am with you on using a finally clause and not introducing a new class unnecessarily though, think I'll redo it that way.

src/main/java/github/scarsz/discordsrv/modules/voice/VoiceModule.java
```diff
@@ -57,44 +58,48 @@ public VoiceModule() {
...
     private void tick() {
-        if (getCategory() == null) {
-            DiscordSRV.debug("Skipping voice module tick, category is null");
-            return;
-        }
-        if (getLobbyChannel() == null) {
-            DiscordSRV.debug("Skipping voice module tick, lobby channel is null");
-            return;
-        }
+        if (!lock.tryLock()) return;
```
Vankka, line 66: Probably should add a debug message when a tick is skipped

PR #678, opened 2020-04-05
Used auto-closing on prepared statements.

metalshark:
Used auto-closing on prepared statements.


Vankka:
Returning inside a try-with-resources statement is ok based on Oracle's standards, https://stackoverflow.com/questions/22947755/try-with-resources-and-return-statements-in-java  
  
PR's good to go though, good job on finding this

PR #681, opened 2020-04-06
Reducing Join/Leave Event Tick Time

metalshark:
I will be running for a while to see if it breaks, but initial testing looks good.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
@@ -154,7 +154,7 @@ public class WebhookUtil {
 
         return lastUsedWebhooks.compute(channel, (c, lastUsedWebhook) -> {
             int index = webhooks.indexOf(lastUsedWebhook);
-            index = index + 1 % webhooks.size();
+            index = (index + 1) % webhooks.size();
             return webhooks.get(index);
         });
     }
--- a/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
@@ -30,7 +30,10 @@ import org.bukkit.entity.Player;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class WebhookUtil {
--- a/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
@@ -68,7 +68,7 @@ public class WebhookUtil {
 
         Bukkit.getScheduler().runTaskAsynchronously(DiscordSRV.getPlugin(), () -> {
             try {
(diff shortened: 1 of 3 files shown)


Vankka:
In terms of latency between the event and the message appearing in Discord? There's not much we can do on that front as most of the latency comes from Cloudflare & Discord itself, though putting that on a seperate thread will certainly have some sort of impact on the amount of time taken on the main thread

metalshark:
> In terms of latency between the event and the message appearing in Discord? There's not much we can do on that front as most of the latency comes from Cloudflare & Discord itself, though putting that on a seperate thread will certainly have some sort of impact on the amount of time taken on the main thread

Wrong terminology - this is reducing in-game latency whilst it processes those events - stops the 100ms+ ticks and gets them back down to 1ms.

src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
```diff
@@ -113,67 +118,45 @@ public static void deliverMessage(TextChannel channel, Player player, String mes
...
-            lastUsedWebhooks.put(channel, new LastWebhookInfo(target.getId(), targetName));
+            return hooks;
+        });
+        if (webhooks == null) return null;
 
-            return target;
-        }
+        return lastUsedWebhooks.compute(channel, (c, lastUsedWebhook) -> {
+            int index = webhooks.indexOf(lastUsedWebhook);
+            index = index + 1 % webhooks.size();
```
Vankka, line 157:
```suggestion
            index = (index + 1) % webhooks.size();
```

src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
```diff
@@ -113,67 +118,45 @@ public static void deliverMessage(TextChannel channel, Player player, String mes
         });
     }
 
-    static class LastWebhookInfo {
-
-        final String webhook;
-        final String targetName;
+    private static final Map<TextChannel, List<Webhook>> channelWebhooks = new ConcurrentHashMap<>();
+    private static final Map<TextChannel, Webhook> lastUsedWebhooks = new ConcurrentHashMap<>();
```
Vankka, line 122: Storing JDA entities is generally a bad idea, they tend to die

metalshark: Oh well, it was fun while it lasted https://github.com/DV8FromTheWorld/JDA/blob/master/README.md#entity-lifetimes

src/main/java/github/scarsz/discordsrv/util/WebhookUtil.java
```diff
@@ -113,70 +119,54 @@ public static void deliverMessage(TextChannel channel, Player player, String mes
         });
     }
 
-    static class LastWebhookInfo {
-
-        final String webhook;
-        final String targetName;
+    private static final Map<String, List<String>> channelWebhookIds = new ConcurrentHashMap<>();
+    private static final Map<String, String> lastUsedWebhookIds = new ConcurrentHashMap<>();
+    private static final int webhookPoolSize = 2;
```
metalshark, line 124: Did you want this exposed in config.yml?

Vankka: 2 is fine, it was initially 2 because of Android (even though that didn't fix the issue), but it's only use would be allowing more bandwidth, iirc the rate limit for webhooks is 5/5, so that would make it 10/5 so 1 message every half a second (iirc)

PR #683, opened 2020-04-08
Added JDBC caching for single server environments

metalshark:
Every use of the account linking lookup resulted in a MySQL query. For a single server environment, caching the results locally seemed more appropriate.


Vankka:
tbh for a single server json or even sqlite or h2 would work fine, I doubt there's many people running mysql and just 1 DiscordSRV instance

metalshark:
I hear you. Any objection to a Redis layer?

Vankka:
DiscordSRV2 (Iris)[https://github.com/DiscordSRV/Iris) should be more of a target for future enhancements & new features, eta on having that is this summer, doubt there will be many more updates on this codebase

PR #686, opened 2020-04-09
VentureChat 2.18.* Updated Plugin Hook

Aust1n46:
Key features:

Utilizes new VentureChatEvent instead of having to manually check for mutes, permissions, etc in the AsyncPlayerChatEvent.

BungeeCord chat channels from across the network can transmit to the same Discord channel with just a single bot and a single installation of DiscordSRV on one Spigot server.

A single bot can transmit chat from Discord to all servers with a single installation of DiscordSRV on one Spigot server.

Multiple DiscordSRV bots and installations can still be used for standard chat channels.

Messages sent from Discord -> Minecraft support VentureChat's message remover via the Moderation GUI.

Caveats: 

The server with DiscordSRV installed on it needs to have at least one player logged on in order to receive and send plugin messages via VentureChat.

Dependency Changes:

Requires VentureChat 2.18.3 (soon to be released)
ProtocolLib

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/hooks/chat/VentureChatHook.java
+++ b/src/main/java/github/scarsz/discordsrv/hooks/chat/VentureChatHook.java
@@ -47,94 +47,79 @@ import java.util.stream.Collectors;
 
 public class VentureChatHook implements ChatHook {
 
-	@EventHandler(priority = EventPriority.NORMAL)
+    @EventHandler(priority = EventPriority.NORMAL)
     public void onVentureChatEvent(VentureChatEvent event) {
-		// event will fire again when received. Don't want to listen twice on the sending server
-		if(event.isBungee()) {
-			return;
-		}
-		
+        // event will fire again when received. Don't want to listen twice on the sending server
+        if (event.isBungee()) return;
+	
         // get channel
         ChatChannel chatChannel = event.getChannel();
-		String channel = chatChannel.getName();
-		
-		String username = event.getUsername();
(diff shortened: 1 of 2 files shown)


src/main/java/github/scarsz/discordsrv/hooks/chat/VentureChatHook.java
```diff
@@ -18,105 +18,155 @@
...
+import net.kyori.text.serializer.legacy.LegacyComponentSerializer;
+
+import org.apache.commons.lang3.StringUtils;
 import org.bukkit.ChatColor;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
-import org.bukkit.event.player.AsyncPlayerChatEvent;
 import org.bukkit.plugin.Plugin;
 
+import com.comphenix.protocol.events.PacketContainer;
```
Vankka, line 43: ProtocolLib would be **required** if this is added, is there a alternative?

Aust1n46: Would only be required if VentureChat is in use, and VentureChat requires it.  I could change around the Format API but any plugin using VentureChat's API will always have access to ProtocolLib since I require it.

Vankka: In that case it'll need to be added to the pom, i'll get ProtocolLib uploaded onto nexus.scarsz.me for this

PR #688, opened 2020-04-11
Redis Support

metalshark:
Adds optional Redis support to the Account Link Manager.


Vankka:
It appears that this is quite tied to the JdbcAccountLinkManager currently, would rather it be in it's own class

metalshark:
> It appears that this is quite tied to the JdbcAccountLinkManager currently, would rather it be in it's own class

It's a caching layer for the database, so quite integral. I can modularise caching to give options between Memcache, Redis and single server in-memory caching.

Vankka:
> caching
ah that makes more sense then, it can stay then; modularity won't be neccasery as there most likely won't be other caching options (especially) in v1

src/main/java/github/scarsz/discordsrv/objects/managers/JdbcAccountLinkManager.java
```diff
@@ -190,6 +209,41 @@ public JdbcAccountLinkManager() throws SQLException {
...
+    private void dropCode(UUID uuid, String code) {
+        final String uuidString = uuid.toString();
+        try (final PreparedStatement statement = connection.prepareStatement("delete from " + codesTable + " where `uuid` = ? and `code` = ?")) {
+            statement.setString(1, uuidString);
+            statement.setString(2, code);
+            statement.executeUpdate();
+            if (jedisPool != null) {
+                try (Jedis jedis = jedisPool.getResource()) {
+                    jedis.del("linking/code/" + code);
+                    jedis.del("linking/uuid/" + uuidString);
```
Vankka, line 241: `Jedis#del`'s accept multiple inputs, eg.
```suggestion
                    jedis.del("linking/code/" + code, "linking/uuid/" + uuidString);
```

src/main/java/github/scarsz/discordsrv/objects/managers/JdbcAccountLinkManager.java
```diff
@@ -201,16 +255,99 @@ private void dropExpiredCodes() {
...
+            } catch (SQLException e) {
+                e.printStackTrace();
+            }
+        }
+        return code;
+    }
+
     @Override
     public Map<String, UUID> getLinkingCodes() {
-        dropExpiredCodes();
```
Vankka, line 206: Removal would cause codes not to be removed other than when the class in constructed

metalshark: I am using
`keyExpiryScheduler.schedule(() -> dropCode(uuid, finalCode), 5, TimeUnit.MINUTES);`
to remove the keys. The cleanup on construction is there just in case of improper shutdown, or shut down prior to the scheduled drop.

Vankka:
We're going to skip adding redis to v1 & consider it again for v2

PR #692, opened 2020-04-17
Use Concurrent Keysets in the Voice Module

metalshark:
Use a concurrent keyset to prevent having to manually declare synchronized sections of code and copy HashSets.


Vankka:
Don't really thing this makes much of a difference

metalshark:
> Don't really thing this makes much of a difference

Was getting some blocking. This should alleviate it with the only downside being that the sets cannot hold null values. The only trouble with extremely rare blocking behaviour is proving its absence. So I cannot prove this works, I can only prove that it doesn't if I still get blocking.

Issue #785, opened 2020-07-07
"Cannot build a Message with more than 2000 characters. Please limit your input."


thecenters:
I get this error as the server is starting up and with no players connected. It's consistent with every reboot. This error causes the console channel to not work. However, the regular chat channel still works.

06.07 00:23:28 [Server] WARN Exception in thread "DiscordSRV - Console Message Queue Worker" java.lang.IllegalStateException: Cannot build a Message with more than 2000 characters. Please limit your input.
06.07 00:23:28 [Server] WARN at github.scarsz.discordsrv.dependencies.jda.api.MessageBuilder.build(MessageBuilder.java:860)
06.07 00:23:28 [Server] WARN at github.scarsz.discordsrv.util.DiscordUtil.queueMessage(DiscordUtil.java:448)
06.07 00:23:28 [Server] WARN at github.scarsz.discordsrv.util.DiscordUtil.sendMessage(DiscordUtil.java:302)
06.07 00:23:28 [Server] WARN at github.scarsz.discordsrv.util.DiscordUtil.sendMessage(DiscordUtil.java:260)
06.07 00:23:28 [Server] WARN at github.scarsz.discordsrv.objects.threads.ConsoleMessageQueueWorker.run(ConsoleMessageQueueWorker.java:47)

Vankka:
Can't reproduce. What version of DiscordSRV are you running?

thecenters:
1.19.1-SNAPSHOT

I have a plugin that spams the console with warnings (it's to be expected and something I can't stop). I'm guessing that's over the character limit and so DiscordSRV crashes on it every time.

Vankka:
DiscordSRV checks & makes sure it's not trying to send anything over 2000 characters. Try updating to the latest development build, https://snapshot.discordsrv.com

thecenters:
Problem persists with latest dev version DiscordSRV-Build-1.19.2-SNAPSHOT-1594412798.jar

`10.07 15:28:44 [Server] WARN Exception in thread "DiscordSRV - Console Message Queue Worker" java.lang.IllegalStateException: Cannot build a Message with more than 2000 characters. Please limit your input.
10.07 15:28:44 [Server] WARN at github.scarsz.discordsrv.dependencies.jda.api.MessageBuilder.build(MessageBuilder.java:860)
10.07 15:28:44 [Server] WARN at github.scarsz.discordsrv.util.DiscordUtil.queueMessage(DiscordUtil.java:448)
[...]

Issue #719, opened 2020-05-29
429 Rate Limits being hit constantly


KsaRedFx:
After running DiscordSRV for... months without issue, at 10pm EST on the 28th our consoles started flooding with messages like this;

```[Fri, 29. May 2020 00:04:42 EDT WARN] [JDA] Encountered 429 on route /channels/:channel_id
[Fri, 29. May 2020 00:04:42 EDT WARN] [JDA] Encountered 429 on route /channels/:channel_id
[Fri, 29. May 2020 00:04:42 EDT WARN] [JDA] Encountered 429 on route /channels/:channel_id
```

Currently running the latest version from the Spigotmc page. 
Unsure why this is happening, perhaps Discord changed their rate limits and for some reason DiscordSRV isn't respecting it? 

Seems to happen for both the console, and chat channel bridges.

KsaRedFx:
The latest snapshot produces... Similar yet slightly different results;

```
[DiscordSRV] [JDA] Encountered 429 on route PATCH/channels/{channel_id} with bucket <SNIP>:guild_id:<SNIP>:webhook_id Retry-After: 147987 ms
```

jasperlachance:
I am having this issue as well. Will await a response from the dev(s).

threlief:
i have same problem lol

KsaRedFx:
I added "Encountered 429 on route" to the list of messages not sent across the Discord Console bridge just to suppress some spam to the console. 

And, the latest snapshot has made the error significantly less frequent.

granny:
Discord appears to be acting up. As long as you are running the latest version (DiscordSRV v1.18.4 or a snapshot 1.19.0 build), we have no control over this. Nothing to do but wait it out.

Network owners (Those that use Bungeecord) can try splitting up their instances so each backend server is operating under their own token, however given how even individual servers are reporting issues, this likely won't fix it.

clarklupine:
Also having this issue, for what it's worth. It's been a really good plugin so far! Appreciate there's not much to be done right now, will attempt KsaRedFx's quick fix of muting the message... as soon as I can figure out how to do that >_<

granny:
That can be done by adding `"Encountered 429"` in the [DiscordConsoleChannelDoNotSendPhrases](https://config.discordsrv.com/config/DiscordConsoleChannelDoNotSendPhrases:) option and running `/discordsrv reload` afterwards.

[...]

PR #799, opened 2020-07-17
Added support for pre/suffixing console messages

SRLcore:
Added support for adding a prefix and suffix to each "batch" of console message lines delivered to Discord. This allows better formatting of the entire batch of messages (eg. allowing users to prefix and suffix the entire batch with `` ``` `` to contain the entire batch in a code block, as pictured below).

Example of code block

This also improves control of the length of the message by adding support for breaking apart console message lines into multiple lines in cases where the single line would itself exceed the maximum message length, possibly fixing #785 . The splitting tries to split on a word boundary if it can, but it will hard-break if the word boundary is too far away. An example of a mock single very long console line being truncated is pictured below.

Example of wrapping a very long console line


src/main/java/github/scarsz/discordsrv/objects/threads/ConsoleMessageQueueWorker.java
```diff
@@ -19,13 +19,25 @@
...
 import net.dv8tion.jda.api.entities.Message;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.text.WordUtils;
+
+import java.util.Deque;
 
 public class ConsoleMessageQueueWorker extends Thread {
 
+    private static final char LINE_WRAP_INDENT = '\t';
+    private static final int MIN_SLEEP_TIME_MILLIS = 1000;
```
Vankka, line 35: This should be higher that the *absolute minimum allowed by Discord*

SRLcore: This was preexisting behavior, my only change here is moving the existing `1000` magic number to a constant. 

JDA should be handling the bulk of actual API rate limiting, assuming it is correctly configured.

What value would you prefer this be changed to?

Vankka: Yes JDA would handle it, but if it would hit a rate limit even once, it would then have one message always waiting in ratelimit limbo to be sent (assuming there's enough things to send every second), I'd suggest 1500 or 2000

SRLcore: Just to be clear, this is a behavioral change to what currently exists in the master/develop branches:
https://github.com/DiscordSRV/DiscordSRV/blob/develop/src/main/java/github/scarsz/discordsrv/objects/threads/ConsoleMessageQueueWorker.java#L61

From what I understand, this is a safety mechanism to prevent the user from setting the value too low in `DiscordConsoleChannelLogRefreshRateInSeconds`. 

Has 1000 ms been found too low? Is there an associated bug with the existing behavior? Of course, I can make this change, I just want to understand the motivation for it and why it has not yet been addressed in the existing implementation.

Vankka: As an example https://github.com/DiscordSRV/DiscordSRV/issues/719 Discord changed the channel update ratelimit to 5 minutes, back then DiscordSRV had a minimum of 5 minutes on the topic updater, yet it still didn't behave properly because it would be a few milliseconds wrong. This hasn't been an issue, atleast that I know of, since the default value fir this is 5 seconds in the config and *most* people don't have flooding console output. We tend not to go with the bare minimum in terms of rate limits, but I just hadn't spotted this until now

SRLcore: No problem. I really appreciate the explanation. I will change it to 2000, since coincidentally that is what I use in my own configuration.

src/main/java/github/scarsz/discordsrv/objects/log4j/ConsoleAppender.java
```diff
@@ -92,7 +92,11 @@ public void append(LogEvent e) {
 
         // return if this is not an okay level to send
         boolean isAnOkayLevel = false;
-        for (String consoleLevel : DiscordSRV.config().getStringList("DiscordConsoleChannelLevels")) if (consoleLevel.toLowerCase().equals(e.getLevel().name().toLowerCase())) isAnOkayLevel = true;
+        for (String consoleLevel : DiscordSRV.config().getStringList("DiscordConsoleChannelLevels")) {
+            if (consoleLevel.toLowerCase().equals(e.getLevel().name().toLowerCase())) {
+                isAnOkayLevel = true;
```
Vankka, line 97: Could break here

PR #802, opened 2020-07-18
Update AccountLinkManager.java

badbrainIRC:
Properly update json file after a discord user link or unlink.


Scarsz:
The file is intentionally not saved after every single account linking action (& the way you've implemented it, it would be saving twice for every link). It wastes time & disk I/O to save every time. Say 500 account pairs are linked, ~30KB of data would be written every single time, even when only one link is changing. It's just not useful.

PR #806, opened 2020-07-21
Added ability to set channel for watchdog

SRLcore:
Currently, the server watchdog message always gets posted in the "main" Discord channel. This adds the ability to configure the destination Discord channel that receives the alert, with a fallback to the main channel if misconfigured or absent.


Scarsz:
Would be much a better implementation to check if a `watchdog` channel pair is defined & use that, defaulting to the main text channel otherwise

SRLcore:
I can add that if required, but I disagree that having a specially-named channel would be better for a few reasons. 

It's generally not good practice to define special values (aka "sentinel values") that change behavior where there would otherwise be none, especially in an externally-facing context. Currently, there are no naming restrictions on any channel in that config, and this change would mean that there is a special value that yields different behavior. This leads to unnecessary cognitive overhead for both users and developers.

There are also potential side-effects for any already-configured channels named `watchdog` that would suddenly start receiving these alerts (perhaps many, due to the recent bugfix allowing more than one batch of alerts to be sent out per server restart), and the only resolution would be to either disable the watchdog entirely, or rename their existing channel to not conflict. Levying this requirement on users of the plugin or API seems unnecessary to me.

In my current implementation, the user is able to select any of their already-defined channels as a destination for the watchdog alerts, and the update does not introduce any behavior changes or side-effects by default. Alternatively, I could follow the same pattern that is used for defining the console channel, with the user specifying the ID directly.

If you would still prefer to have a special channel name for watchdog alerts, I will make that change - I just wanted to explain why I disagree with it being "a much better implementation."

Scarsz:
> It's generally not good practice to define special values (aka "sentinel values") that change behavior where there would otherwise be none, especially in an externally-facing context. Currently, there are no naming restrictions on any channel in that config, and this change would mean that there is a special value that yields different behavior. This leads to unnecessary cognitive overhead for both users and developers.

While I get what you're saying, it's not a sentinel value as you described. Reserving channel names for special purposes is surely not a problem, especially considering the alerts feature is brand new & the term `watchdog` is already a well understood term with server software.

> There are also potential side-effects for any already-configured channels named `watchdog` that would suddenly start receiving these alerts (perhaps many, due to the recent bugfix allowing more than one batch of alerts to be sent out per server restart), and the only resolution would be to either disable the watchdog entirely, or rename their existing channel to not conflict. Levying this requirement on users of the plugin or API seems unnecessary to me.

As I said previously, I don't see it as a problem at all. As long as it's documented that this behavior existed, there shouldn't be any issues. Not enough people use alerts for a `watchdog` channel name to happen as it stands & I'm sure nobody's using the same channel name for anything special already.

> In my current implementation, the user is able to select any of their already-defined channels as a destination for the watchdog alerts, and the update does not introduce any behavior changes or side-effects by default. Alternatively, I could follow the same pattern that is used for defining the console channel, with the user specifying the ID directly.

It's redundant is the thing. DiscordSRV has a channel system for a reason, going around it isn't useful. The console channel is going to be switched to the same system, defining a `console` channel.

SRLcore:
Updated according to your request.

SRLcore:
I've found an even better reason for why this is not a good idea. 

The channel mapping is enforced to be 1:1 by the existence of the `getDestinationGameChannelNameForTextChannel` and `getDestinationTextChannelForGameChannelName` methods. My use case is that I want the alerts to go to a "staff" channel, that aggregates alerts from multiple plugins. This change would force me to change the name of my existing "staff" channel (called `staff`) to `watchdog`, and propagate the updates to all of the plugins where this needs to be configured.

Let me know if this changes your position.

Scarsz:
> The channel mapping is enforced to be 1:1 by the existence of the `getDestinationGameChannelNameForTextChannel` and `getDestinationTextChannelForGameChannelName` methods.

That isn't true. While the channel names *do* need to be unique, the target channel IDs on Discord do not.

> My use case is that I want the alerts to go to a "staff" channel, that aggregates alerts from multiple plugins. This change would force me to change the name of my existing "staff" channel (called `staff`) to `watchdog`, and propagate the updates to all of the plugins where this needs to be configured.

The following is perfectly valid and will be considered the correct way to accomplish this in the future:

`{"staff-chat": "01234567890012345", "watchdog": "01234567890012345"}`

src/main/java/github/scarsz/discordsrv/objects/threads/ServerWatchdog.java
```diff
@@ -66,7 +68,10 @@ public void run() {
                         continue;
                     }
 
-                    String channelName = DiscordSRV.getPlugin().getMainTextChannel().getName();
+                    @SuppressWarnings("ConstantConditions" /* getDestinationTextChannelForGameChannelName may return null */)
```
Vankka, line 71: Since there is only one suppression it'd look neater to have it like this (also matches existing code style better):
```java
@SuppressWarnings("ConstantConditions") // comment
```

SRLcore: I will make this change to comply with your style guidelines. The reason I prefer the comment inside of the annotation, is that my preferred IDE (IntelliJ IDEA), will collapse the comment as pictured in the attachment, resulting in a neater view. Clicking on the collapsed annotation expands it and shows the comment when necessary.

Comment inside:
image

Comment outside:
image

Issue #776, opened 2020-06-30
Poor performance on main thread with MySQL linked account backend + LuckPerms context


Vankka:
https://i.gyazo.com/76c8fbb14e4d1d94fee86d8d96b3f089.png
https://i.gyazo.com/thumb/1200/b92015e806a4f434f4f868ca7358320c-png.jpg

Caused by the use of MySQL connections on the main thread & LuckPerms determining contexts

Vankka:
Fixed in development builds, https://snapshot.discordsrv.com

PR #865, opened 2020-09-19
Overhauled internal account linking implementations

Scarsz:


Vankka (changes requested):
Here's some general comments;
- We may want to store if a player has already been linked once, so if they unlink & link again we can have a option to not give out (specific) rewards
- I'm guessing there's probably a plan to for doing as little mysql/http on the main thread, but if there isn't there probably should be (https://github.com/DiscordSRV/DiscordSRV/issues/776)

also left a few inline comments

Scarsz:
> * I'm guessing there's probably a plan to for doing as little mysql/http on the main thread, but if there isn't there probably should be (#776)

Went ahead and added some async methods to the providers https://github.com/DiscordSRV/DiscordSRV/pull/865/commits/21a8661bc99d9ba68d6ab5d26b266a4428696c59

Vankka:
> Went ahead and added some async methods to the providers 21a8661

In the instance of https://github.com/DiscordSRV/DiscordSRV/issues/776, LuckPerms' ContextCalculator this wouldn't work, see javadoc comment: https://github.com/lucko/LuckPerms/blob/2ea43d4b2ee63aa13eac63c6d6dd1d0fc7c1fd15/api/src/main/java/net/luckperms/api/context/ContextCalculator.java#L39-L43

Scarsz:
> In the instance of #776, LuckPerms' ContextCalculator this wouldn't work, see javadoc comment: https://github.com/lucko/LuckPerms/blob/2ea43d4b2ee63aa13eac63c6d6dd1d0fc7c1fd15/api/src/main/java/net/luckperms/api/context/ContextCalculator.java#L39-L43

I feel that's an issue best handled inside the `ContextCalculator` itself, the providers being async should be enough to cover the feature set of a provider

Maybe there could be a `CachedAccountSystem` subclass to implement that functionality as well

Issue #890, opened 2020-10-24
Advancements don't match up


bitttttten:
In-game I got the two advancements Those Were the Days and War Pigs, although in the discord channel the bot reported that I got "Find Bastion" and "Loot Bastion" respectively. It seems that none of the other advancement titles match up either.

GreemDev:
DiscordSRV uses the advancement's internal name. "Those Were The Days" is obtained from finding a Bastion, and War Pigs is obtained from looting a Bastion. That's what they're called internally so we use it, though I don't remember exactly why we don't use the in-game names for advancements but I believe it had something to do with i18n.

underscore11code:
>I don't remember exactly why we don't use the in-game names for advancements but I believe it had something to do with i18n.

iirc advancement names are handled clientside for localization, so the server never knows what the "pretty" name is.

GreemDev:
> 
> 
> > I don't remember exactly why we don't use the in-game names for advancements but I believe it had something to do with i18n.
> 
> iirc advancement names are handled clientside for localization, so the server never knows what the "pretty" name is.

Ah, that's it.

bitttttten:
Makes sense. Thanks all!

PR #915, opened 2020-11-28
Make advancement messages use localized name

bmxguy100:
Makes the advancement messages try to use the localized names from en_us.json before resorting to the name used internally in minecraft's advancement json files. Fixes DiscordSRV/DiscordSRV#890.


Scarsz:
Have you confirmed that this implementation works as expected?

bmxguy100:
I have tested it for a few advancements on 1.16.4, and will test some other versions soon.

bmxguy100:
I have confirmed that it also works with 1.12.2

Issue #791, opened 2020-07-10
Allow one way group synchronization


Jozznak:
Hello, 
For security reasons, it would be very useful if you make possible to disable Discord as authoritative. In this way we only allow Minecraft Sync with Discord roles and they can only be managed through MC. This prevents a lot of possible ways we can be fucked up, but to resume I'll point 2 things: 

First, if a staff get hacked or something, In our mc server, roles can only be managed through the console. But in Discord, as we have admin role, we can actually modify roles, so if a staff gets hacked (It really happens) the hacker could make a real mess and fuck us up.
 
Second and most probable thing to happen. We never really finish to know people. And it could be that a staff decides to fuck up the server giving ranks to all people through discord before leaving or what ever reason he has. Of course, this 2 aspects are not DiscordSRV fault, but if you add this option, it could really help us to prevent this and more things that could happen. Human errors and stupidity... Yeah, we have trust issues now days. And that's why we limit so much how roles are administrated in game. If something happens in Discord with the roles, is not really that bad as there is no money involved, no special items or cosmetics (at least in our case). So it can be easily fixed. But if this happens in our mc server, it would break a lot of things.

So I hope this can be integrated, because it would be really useful for does who have the same concerns than us.
Thanks.

Vankka:
You *could*, put the wrong group names in the config and then grant the actual groups the permission `discordsrv.sync.<wrong group name>`, then group sync would effectively become one-way, thought that's obviously not the cleanest solution

Jozznak:
I will try that, but is there a posibility you take this in consideration?

Vankka:
Considering that there is a valid solution for now, no. In the future there will likely be a way to configure if specific synchronization pairs should be one way

Vankka:
Available on development builds, https://snapshot.discordsrv.com

PR #918, opened 2020-11-29
One-way group/role synchronization

jvdoorn:
In light of issue #791 I made `GroupRoleSynchronizationMinecraftIsAuthoritative` strictly authoritative (even when synchronising to MineCraft). If preferred I can make a separate config option for this so that `GroupRoleSynchronizationMinecraftIsAuthoritative` maintains its current behaviour. I also added a debug statement that might help to debug any issues related to authoritativeness in the future.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
+++ b/src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
@@ -182,6 +182,7 @@ public class GroupSynchronizationManager extends ListenerAdapter implements List
         boolean oneWaySynchronisation = DiscordSRV.config().getBoolean("GroupRoleSynchronizationOneWay");
         boolean minecraftIsStrictlyAuthoritative = oneWaySynchronisation && DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative");
         boolean discordIsStrictlyAuthoritative = oneWaySynchronisation && !DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative");
+        synchronizationSummary.add("Synchronisation is one way (" + (minecraftIsStrictlyAuthoritative ? "MineCraft -> Discord" : "Discord -> MineCraft") + ")");
 
         for (Map.Entry<String, String> entry : DiscordSRV.getPlugin().getGroupSynchronizables().entrySet()) {
             String groupName = entry.getKey();
@@ -267,7 +268,6 @@ public class GroupSynchronizationManager extends ListenerAdapter implements List
                 || (!roleIsManaged
                     && !discordIsStrictlyAuthoritative
                     && (direction == SyncDirection.AUTHORITATIVE ? DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative") : direction == SyncDirection.TO_DISCORD));
-            DiscordSRV.debug((minecraftIsAuthoritative ? "MineCraft" : "Discord") + "is authoritative for synchronising {" + groupName + ":" + role + "} for {" + player.getName() + ":" + user + "}");
 
             if (hasGroup == hasRole) {
                 // both sides agree, no changes necessary
--- a/src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
+++ b/src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
@@ -179,12 +179,9 @@ public class GroupSynchronizationManager extends ListenerAdapter implements List
         Map<Guild, Map<String, Set<Role>>> roleChanges = new HashMap<>();
 
         // Check if Minecraft or Discord is strictly authoritative.
-        boolean minecraftIsStrictlyAuthoritative = DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsStrictlyAuthoritative");
-        boolean discordIsStrictlyAuthoritative = DiscordSRV.config().getBoolean("GroupRoleSynchronizationDiscordIsStrictlyAuthoritative");
-        if (minecraftIsStrictlyAuthoritative && discordIsStrictlyAuthoritative) {
-            DiscordSRV.debug("Both Minecraft and Discord are set to be strictly authoritative, Minecraft will take precedence");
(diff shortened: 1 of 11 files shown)


Vankka:
This would prevent manual Discord role changes from modifying Minecraft groups (when Minecraft is authoritative).
I'd rather see a second config option that makes the synchronization one way (both for Minecraft and Discord, rather than just Minecraft, too)

jvdoorn:
@Vankka I updated my solution, is this more to your liking? It adds two new config variables: `GroupRoleSynchronizationMinecraftIsStrictlyAuthoritative` and `GroupRoleSynchronizationDiscordIsStrictlyAuthoritative`. If both are set to `true` then MineCraft will take precedence. This will be printed to the debug. 

I do not know how/if the default config files need to be updated. So that is something that might still need to be done before merging this PR.

Vankka:
What I was in visioning is like a `GroupRoleSynchronizationOneWay`, and then will decide which way from `GroupRoleSynchronizationMinecraftIsAuthoritative`

> I do not know how/if the default config files need to be updated. So that is something that might still need to be done before merging this PR.

All 10 of the synchronization files would need new options to be added, comments don't need to be translated since synchronization files aren't translated

jvdoorn:
Done so, @Vankka!

src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
```diff
@@ -253,9 +258,16 @@ public void resync(OfflinePlayer player, SyncDirection direction, boolean addLin
...
+            if (roleIsManaged && minecraftIsStrictlyAuthoritative) {
+                synchronizationSummary.add("Tried to sync {" + role + ":" + groupName + "} to Discord but the role is managed and Minecraft is strictly authoritative");
+                continue;
+            }
+            // Determine if Minecraft is authoritative in the synchronization.
+            boolean minecraftIsAuthoritative = minecraftIsStrictlyAuthoritative
+                || (!roleIsManaged
+                    && !discordIsStrictlyAuthoritative
+                    && (direction == SyncDirection.AUTHORITATIVE ? DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative") : direction == SyncDirection.TO_DISCORD));
+            DiscordSRV.debug((minecraftIsAuthoritative ? "MineCraft" : "Discord") + "is authoritative for synchronising {" + groupName + ":" + role + "} for {" + player.getName() + ":" + user + "}");
```
Vankka, line 270: Leave this debug out (see previous comment)

src/main/resources/synchronization/en.yml
```diff
@@ -20,6 +20,8 @@ NicknameSynchronizationFormat: "%displayname%"
 #                                               {"MC_GROUP_NAME": "DISCORD_ROLE_ID"} is the format to go by when adding more group/role pairs
 #                                               to get your Discord guild's role IDs, run "/discord debug" and look at the first section
 # GroupRoleSynchronizationMinecraftIsAuthoritative: whether or not Minecraft group changes override Discord role changes
+# GroupRoleSynchronizationOneWay: whether or not to synchronise only one way, the way it is synchronised depends on the value
+#                                 of GroupRoleSynchronizationMinecraftIsAuthoritative.
```
Vankka, line 24: This comment should be added to all the files (but doesn't need to be translated)

src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
```diff
@@ -178,6 +178,11 @@ public void resync(OfflinePlayer player, SyncDirection direction, boolean addLin
 
         Map<Guild, Map<String, Set<Role>>> roleChanges = new HashMap<>();
 
+        // Check if Minecraft or Discord is strictly authoritative.
+        boolean oneWaySynchronisation = DiscordSRV.config().getBoolean("GroupRoleSynchronizationOneWay");
+        boolean minecraftIsStrictlyAuthoritative = oneWaySynchronisation && DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative");
+        boolean discordIsStrictlyAuthoritative = oneWaySynchronisation && !DiscordSRV.config().getBoolean("GroupRoleSynchronizationMinecraftIsAuthoritative");
```
Vankka, line 184: Should log these to `synchronizationSummary`, eg. `Synchronization is one way to Minecraft`

Vankka (commented):
This looks great, left a few inline comments for a few changes

src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
jvdoorn: Done!

src/main/resources/synchronization/en.yml
jvdoorn: Done, I translated the Dutch one.

src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
jvdoorn: Done!

src/main/java/github/scarsz/discordsrv/objects/managers/GroupSynchronizationManager.java
```diff
@@ -178,6 +178,12 @@ public void resync(OfflinePlayer player, SyncDirection direction, boolean addLin
...
+        synchronizationSummary.add("Synchronisation is one way (" + (minecraftIsStrictlyAuthoritative ? "MineCraft -> Discord" : "Discord -> MineCraft") + ")");
```
Scarsz, line 185: Use proper capitalization of `Minecraft`

Vankka, line 185: This should only be added if the synchronization *is* one way

jvdoorn: Good catch!

jvdoorn: Done!

Vankka (approved):

PR #960, opened 2021-01-06
Exclude gson and junit from shading. Relocate some packages which were not relocated yet.

Janmm14:
Reduces file size slightly.

It is possible that the inclusion of not-relocated gson is the cause of
https://hub.spigotmc.org/jira/browse/SPIGOT-6056?focusedCommentId=37758&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-37758
(also related to https://github.com/konsolas/AAC-Issues/issues/2108 )

Also see https://github.com/Scarsz/Configuralize/pull/2


Vankka:
1.7, which we do still support for modded (bungee + forge hack) servers doesn't include gson as far as I know, I think it'd be better to relocate it instead

Androkai:
> 1.7, which we do still support for modded (bungee + forge hack) servers doesn't include gson as far as I know, I think it'd be better to relocate it instead

Is missing in 1.7.10 so agree with relocation.
image

Janmm14:
added gson again and shading it

Vankka (approved):

PR #995, opened 2021-01-25
Add SuperVanish join/leave hook

goweiwen:
Adds a dependency due to not using reflection, but I think it should be fine since this is how the chat hooks are implemented too.

- Respects `discordsrv.silentjoin` and `discordsrv.silentquit`.
- Respect SuperVanish's silent setting
- Uses join/leave message set in DiscordSRV config

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -358,6 +358,10 @@
             <id>sonatype</id>
             <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
         </repository>
+        <repository>
+            <id>jitpack.io</id>
+            <url>https://jitpack.io</url>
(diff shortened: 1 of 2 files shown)


Vankka:
This ignores SuperVanish' config options, https://github.com/LeonMangler/SuperVanish/blob/9c4a6046b8590d359765cbb23939e7a13dd1fbca/src/main/java/de/myzelyam/supervanish/features/Broadcast.java#L73

goweiwen:
Fixed

Scarsz (approved):

Vankka (approved):

PR #1005, opened 2021-01-30
Korean translation

ShapeLayer:


Vankka (changes requested):
You've completely changed the formatting of the Korean files, making them inconsistent with the others,
- There should not be more than one empty line of gap between options
- The comments should be in the same place as the English files, specifically they should not be split for every option
- You've added semicolons (`:`) when breaking a line, this seems quite unconventional

ShapeLayer:
Oh, I'm sorry very much.
As focusing on Korean translation and obvious meaning, I neglected the original format.
I'll fix it as soon as possible.

ShapeLayer:
I worked done. Please check the code.

granny (approved):

PR #1050, opened 2021-03-11
Change PlayerQuitEvent priority to LOWEST

Dinty1:
VanishNoPacket's leave listener is at priority `NORMAL`, and DiscordSRV needs to receive this event before so that it is actually able to see if the player is vanished (the player is removed from the vanish list after they quit).

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/PlayerJoinLeaveListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/PlayerJoinLeaveListener.java
@@ -102,7 +102,7 @@ public class PlayerJoinLeaveListener implements Listener {
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/github/scarsz/discordsrv/listeners/PlayerJoinLeaveListener.java
```diff
@@ -102,7 +102,7 @@ public void onPlayerJoin(PlayerJoinEvent event) {
         }
     }
 
-    @EventHandler //priority needs to be different to MONITOR to avoid problems with permissions check when PEX is used
+    @EventHandler(priority = EventPriority.LOWEST) //priority needs to be different to MONITOR to avoid problems with permissions check when PEX is used, is at lowest so that it executes before VanishNoPacket's player leave listener and is able to see whether the player is vanished
```
Vankka, line 105:
```suggestion
    @EventHandler(priority = EventPriority.LOWEST) //priority needs to be different to MONITOR to avoid problems with permissions check when PEX is used, it is at lowest so that it executes before VanishNoPacket's player leave listener and is able to see whether the player is vanished
```

Vankka (approved):

PR #1070, opened 2021-04-07
Show/hide/name options on emojis/emotes

Scarsz:


src/main/java/github/scarsz/discordsrv/objects/DiscordSRVMinecraftRenderer.java
```diff
@@ -54,7 +54,8 @@
 
     @Override
     public @Nullable Component appendEmoteMention(@NonNull Component component, @NonNull String name, @NonNull String id) {
```
Vankka, line 56: This is only in use when reserialization is used

src/main/resources/config/et.yml
```diff
@@ -110,6 +110,8 @@ DiscordChatChannelMinecraftToDiscord: true
 DiscordChatChannelTruncateLength: 256
 DiscordChatChannelTranslateMentions: true
 DiscordChatChannelAllowedMentions: [user, channel, emote]
+DiscordChatChannelEmojiBehavior: "name"
+DiscordChatChannelEmoteBehavior: "name"
```
Scarsz, line 114: Comments

Vankka (approved):

PR #1072, opened 2021-04-08
Undelete Estonian

Madis0:
This reverts commit df9db87dc130398e5eb430cfe0190335fa58e771.

The point of the Estonian translation in DiscordSRV was to _translate every user-facing message, while keeping admin options/comments as-is_. So admins can simply get DiscordSRV, configure it to their needs and change the language, without having to worry about translating individual phrases that the users will see.
This is in line with other plugins that usually don't let people translate config messages at all, as admins very likely know English enough to install and configure it anyway. Plus, any help channels such as the wiki here are written in English, so it helps to know the original phrasing.

Is there a problem with this approach or do you suggest a better way to do it?


Scarsz:
My bad, should've checked back with why they were added in the first place.

The `alerts.yml` & `linking.yml` files don't need to have an `et` version, though, because they don't have any translated text

Scarsz:
I just had a look at how DiscordSRV handles missing files in a translation and discovered it will refuse to use the language if some are missing. Sorry about that.

Just go ahead and add those back & I'll merge

src/main/resources/messages/en.yml
```diff
@@ -6,8 +6,8 @@
 # You can specify a different format per channel. Let's say you have a channel named: "mychannel".
 # If you want it to have other formatting than the one defined globally you can add following properties:
 #
-# DiscordToMinecraftChatMessageFormat_mychannel: "[&bDiscord From MyChannel &r| %toprolecolor%%toprole%&r] %name% » %message%"
-# DiscordToMinecraftChatMessageFormatNoRole_mychannel: "[&bDiscord From MyChannel&r] %name% » %message%"
+# DiscordToMinecraftChatMessageFormat_mychannel: "[&bDiscord kanalilt Minukanal &r| %toprolecolor%%toprole%&r] %name% » %message%"
+# DiscordToMinecraftChatMessageFormatNoRole_mychannel: "[&bDiscord kanalilt Minukanal&r] %name% » %message%"
```
Scarsz, line 10: You put this in the wrong file

Scarsz (changes requested):

Scarsz (approved):

PR #1079, opened 2021-04-11
Specify that the JDBC connection string error is safe to ignore if the user is not trying to use JDBC

Dinty1:
Recently in #support, a number of people have posted this error and asked if it was something to worry about/how they should go about fixing it. This aims to tell them that it is something which is safe to ignore provided they are not trying to use JDBC.


Vankka (changes requested):
I don't like this, the reason it's erroring is because the pattern fails to match with the default value because the port is set to `PORT` instead of a number - I'd rather the skip be skipped entirely when the option is set to the default value

Dinty1:
Fixed by Scarsz.

Issue #1030, opened 2021-02-19
Suggestion: Add discord Name variable into nickname sync


Tringlle:
It would be really cool if I could do something like this with name sync `[MC NAME] - [DISCORD NAME]`

Dinty1:
This is currently available on the [latest snapshot](https://snapshot.discordsrv.com) and will be added in the next stable release.

PR #1098, opened 2021-04-23
Nickname Sync Improvements

### LINKED ISSUE #506: Permission for minecraft name sync
Add a permission to bypass the name sync (mc to discord) or make it default and we can negate it for certain groups / users.

### LINKED ISSUE #1093: Add support for more placeholders in nickname synchronization
Currently the only placeholders available for nickname synchronization in synchronization.yml are %displayname% and %username% which are both from minecraft.

A feature to add would be to allow the synching to include placeholders like the discord username or other things from PAPI.

Dinty1:
- Adds `%discord_name%` and `%discord_discriminator%` placeholders to the `NicknameSynchronizationFormat` config option (closes #1093 and #1030)
 - Adds a `discordsrv.nicknamesync` permission, which is set to true by default but can be negated if admins don't want a specific group to have names synced (closes #506)

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
+++ b/src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
@@ -131,7 +131,7 @@ public class NicknameUpdater extends Thread {
 
         nickname = MessageUtil.strip(nickname);
         if (nickname.length() > 32) {
-            DiscordSRV.debug("The new nickname for " + offlinePlayer.getName() + " is too long, reducing it to 32 characters.");
+            DiscordSRV.debug("The new nickname for " + offlinePlayer.getName() + " (" + nickname + ") is too long, reducing it to 32 characters.");
             nickname = nickname.substring(0, 32);
         }
         DiscordUtil.setNickname(member, nickname);
--- a/src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
+++ b/src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
@@ -131,7 +131,7 @@ public class NicknameUpdater extends Thread {
 
         nickname = MessageUtil.strip(nickname);
         if (nickname.length() > 32) {
-            DiscordSRV.debug("The new nickname for " + offlinePlayer.getName() + " is too long, reducing it to 32 characters.");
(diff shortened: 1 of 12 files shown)


src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
```diff
@@ -117,7 +120,9 @@ public void setNickname(Member member, OfflinePlayer offlinePlayer) {
 
             nickname = DiscordSRV.config().getString("NicknameSynchronizationFormat")
                     .replace("%displayname%", player.getDisplayName() != null ? player.getDisplayName() : player.getName())
-                    .replace("%username%", player.getName());
+                    .replace("%username%", player.getName())
+                    .replace("%discord_name%", member.getUser().getName())
+                    .replace("%discord_discriminator%", member.getUser().getDiscriminator());
```
Vankka, line 125: Now that these are included, somebody might try to do something like `%username% | %discord_name%` which would lead to the nickname being too long (>32 characters), it'd be worth checking (either at the end of this block or in DiscordUtil) to see if the nickname is too long, and if it is substringing it to size

Dinty1: I'll look into doing that

Dinty1: Added

src/main/java/github/scarsz/discordsrv/objects/threads/NicknameUpdater.java
```diff
@@ -117,14 +120,20 @@ public void setNickname(Member member, OfflinePlayer offlinePlayer) {
...
 
             nickname = PlaceholderUtil.replacePlaceholders(nickname, player);
         } else {
             nickname = offlinePlayer.getName();
         }
 
         nickname = MessageUtil.strip(nickname);
+        if (nickname.length() > 32) {
+            DiscordSRV.error("The new nickname for " + offlinePlayer.getName() + " is too long, reducing it to 32 characters.");
```
Vankka, line 134: As I already stated on Discord, would like to see someway to avoid this showing up as a error, incase the user is ok with them being shortened

Vankka (approved):

PR #1114, opened 2021-05-08
Add support for Discord replies

### LINKED ISSUE #1029: Discord Replies support
sometimes players of my server gets confused when someone on discord replies to someone else, cause it just shows what they said and not the fact that they were replying to someone... i feel like it should be a feature so that people would know your replying to someone without checking the discord. heres an example of how i feel it should work
unknown

Dinty1:
Closes #1029 

Marking as draft since I'm not 100% sure about the config wording so I don't want to translate it yet.

Feedback on said wording would be appreciated :)


src/main/resources/messages/en.yml
```diff
@@ -28,12 +28,24 @@
 #                    example: Notch
 # %channelname%:    name of the channel that the message is coming from
 #                    example: server-chat
+# %reply%:          the message displayed when the message is a reply to another message
+#                    message format can be configured with DiscordToMinecraftMessageReplyFormat
+#                    is not included if the message isn't a reply
```
Vankka, line 33: Has a grammatical error and is somewhat unclear, I'd suggest something like the below
```suggestion
# %reply%:          the message displayed when the message is a reply to another message.
#                    The message format can be configured with DiscordToMinecraftMessageReplyFormat,
#                    this will be empty if the message isn't a reply to another message
```

src/main/resources/messages/en.yml
```diff
@@ -28,12 +28,24 @@
...
 #
 # DiscordToMinecraftAllRolesSeparator: the separator used in between roles in %allroles%
 #
-DiscordToMinecraftChatMessageFormat: "[<aqua>Discord</aqua> | %toprolecolor%%toprolealias%<reset>] %name% » %message%"
-DiscordToMinecraftChatMessageFormatNoRole: "[<aqua>Discord</aqua>] %name% » %message%"
+# DiscordToMinecraftMessageReplyFormat: the format for the message displayed to indicate that the message is a reply
+#
+# Available placeholders:
+# %name%:          effective name on Discord (nickname if present, username otherwise) for the user who is being replied to
```
Vankka, line 40:
```suggestion
# %name%:          the effective name of the user that is being replied to on Discord (nickname if present, username otherwise)
```

src/main/resources/messages/en.yml
```diff
@@ -28,12 +28,24 @@
...
+#                   example: NotchIsMe
+# %username%:      username on Discord for the user who is being replied to
```
Vankka, line 42:
```suggestion
# %username%:      the username of the user that is being replied to on Discord
```

src/main/resources/messages/en.yml
```diff
@@ -28,12 +28,24 @@
...
+#                   example: Notch
+#
```
Vankka, line 44: A placeholder for the replied message is missing

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -304,7 +304,20 @@ private String replacePlaceholders(String input, GuildMessageReceivedEvent event
...
+    private String replaceReplyPlaceholders(String format, Message repliedMessage) {
+        Function<String, String> escape = MessageUtil.isLegacy(format)
+                ? str -> str
+                : str -> str.replaceAll("([<>])", "\\\\$1");
+
+        return format.replace("%name%", escape.apply(MessageUtil.strip(repliedMessage.getMember().getEffectiveName())))
+                .replace("%username%", escape.apply(MessageUtil.strip(repliedMessage.getAuthor().getName())))
+                .replace("\\~", "~") // get rid of escaped characters, since Minecraft doesn't use markdown
+                .replace("\\*", "") // get rid of escaped characters, since Minecraft doesn't use markdown
+                .replace("\\_", "_"); // get rid of escaped characters, since Minecraft doesn't use markdown
```
Vankka, line 320: A lot of this is duplicated from the above method, look into reducing that, for example the need to remove shitty escaping of markdown can be removed if the %reply% replacement is done higher up in the above method.

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -304,7 +304,20 @@ private String replacePlaceholders(String input, GuildMessageReceivedEvent event
                 .replace("\\~", "~") // get rid of escaped characters, since Minecraft doesn't use markdown
                 .replace("\\*", "") // get rid of escaped characters, since Minecraft doesn't use markdown
                 .replace("\\_", "_") // get rid of escaped characters, since Minecraft doesn't use markdown
-                .replace("%message%", message);
+                .replace("%message%", message)
+                .replace("%reply%", event.getMessage().getReferencedMessage() != null ? replaceReplyPlaceholders(DiscordSRV.config().getString("DiscordToMinecraftMessageReplyFormat"), event.getMessage().getReferencedMessage()) : "");
```
Vankka, line 308: Could (or probably should) use a LangUtil.Message instead of DiscordSRV.config().getString()

src/main/resources/messages/en.yml
Vankka: Hovering to show things is possible with MiniMessage, https://docs.adventure.kyori.net/minimessage.html#hover

Dinty1: TIL. I'll add that to the default value as well

Dinty1: I've added the placeholder and made the message display on hover, but since this will only work with non-legacy DiscordToMinecraftChatMessageFormat values I think it might be worth saying as much somewhere so that we don't get the inevitable "help the hover isn't working"

Alternatively we could just leave this out of the default value

Vankka: Those options will not get updated with %reply% automatically, either. Arguably the available formats should be documented way better, but it is what ut is

Dinty1: Added a comment to tell users that they must use MiniMessage formatting for the hover text to work, though I think somehow converting these values from legacy to MiniMessage would be a better way of doing it.

Dinty1:
Everything has been cleaned up and translated so I've marked this as ready for review

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -301,12 +301,23 @@ private String replacePlaceholders(String input, GuildMessageReceivedEvent event
...
                 .replace("\\_", "_") // get rid of escaped characters, since Minecraft doesn't use markdown
                 .replace("%message%", message);
     }
 
+    private String replaceReplyPlaceholders(String format, Message repliedMessage) {
+        Function<String, String> escape = MessageUtil.isLegacy(format)
+                ? str -> str
+                : str -> str.replaceAll("([<>])", "\\\\$1");
+
+        return format.replace("%name%", escape.apply(MessageUtil.strip(repliedMessage.getMember().getEffectiveName())))
```
Vankka, line 316: `getMember` is most likely null if the user someone replied to left the server

src/main/resources/messages/en.yml
```diff
@@ -28,12 +28,27 @@
...
+# DiscordToMinecraftMessageReplyFormat: the format for the message displayed to indicate that the message is a reply
+#
+# Available placeholders:
+# %name%:          the effective name of the user that is being replied to on Discord (nickname if present, username otherwise)
+#                   example: NotchIsMe
+# %username%:      the username of the user that is being replied to on Discord
+#                   example: Notch
+# %message%:       the content of the message that is being replied to
+#
+# NOTE: The hover text will only work if your Discord to Minecraft chat format uses MiniMessage formatting (https://docs.adventure.kyori.net/minimessage.html#the-components)
```
Vankka, line 46: This should probably also include that %reply% must be present in the format

Dinty1:
The new config options need to be added to the Polish messages.yml. I'll do this tomorrow.

Vankka (approved):

Dinty1:
Found a bug which may or may not be caused by this. Will investigate so don't merge yet please :)

Dinty1:
Can't seem to reproduce the issue after more testing so It's probably nothing to worry about

granny (approved):

PR #1137, opened 2021-05-26
Remove alerts

Vankka:
Waiting on https://github.com/DiscordSRV/Alerts to be finalized


src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -406,6 +402,10 @@ public DiscordSRV() {
                     .findFirst().ifPresent(config::setLanguage);
         }
 
+        if (new File(getDataFolder(), "alerts.yml").exists()) {
+            // TODO: log message stating alerts is now a separate plugin
```
Vankka, line 406: Leaving a comment here to remember this

Scarsz:
I want to keep DiscordSRV's alerts system in place for now.

When the dedicated Alerts plugin is released, we can defer people to use that one if they want more advanced functionality than DiscordSRV provides

Issue #1089, opened 2021-04-16
Deprecated JDBC driver warning


mindgam3s:
Console displays
```
[16:12:02] [DiscordSRV - Initialization/WARN]: Loading class 'com.mysql.jdbc.Driver'. This is deprecated. The new driver class is 'com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary.
```

git-Paper-584 (MC: 1.16.5)

https://bin.scarsz.me/087ff1a9-a4ce-466a-9f92-5a9ce90852ca#F4HqnXDIBQUyWA92LlISsdjxwSUFj3ic

PlayingWithPi:
Seems like this is an ongoing issue, wondering if anyone is working on it.
git-Paper-187 (MC: 1.18.1)

SatoriHoshiAiko:
I investigated this and resolved the source of this warning.

In my case to replicate the issue, I was using MariaDB and not mysql.

The warning happens actually more frequently when the driver fails to load. I.e. a bad syntax in the configuration such as jdbc:http:/localhost:3306/my_db - In this case missing a :// for :/ - The driver won't load from bad syntax.

But this same issue will happen when you use mariadb. I tried passing jbdc:mariadb://localhost:3306/mydb in the config - In this case, it loaded correctly (because I had mariadb installed) but DiscordSRV noted that mariadb was unsupported and so it would skip.

To resolve - Make sure you are using Mysql (mysql-community-server e.g.) - Follow traditional guidelines to install mysql and have it working. If you need to migrate from mariadb to mysql, be forewarned, you will have to be careful about backing up your databases before switching, **You can lose your database data!!**


In my case, I had no existing databases I didn't care to lose - And so I uninstalled mariadb, installed mysql. Using jbdc:mysql://localhost:3306/mydb - It now works, it does not fail to load the driver because I am using mysql and not mariadb.

For anyone who care to know how to resolve this.

Also this "bug" - It is a warning and in fact it will be follow in some form by Cannot make connection. Well indeed it can't, DiscordSRV JBDC is not even using mariadb as of yet it would seem. How can is load a mysql driver when it runs mariadb? They are similar but not same. But a working Mysql server and correct database credentials? Yes it will work. Double check your syntax. Double check your login credentials. Backup your database if you have to migrate.

[...]

PR #1150, opened 2021-06-07
Overhauled internal account linking implementations

Scarsz:


src/main/java/github/scarsz/discordsrv/objects/managers/AccountLinkManager.java
```diff
@@ -1,157 +0,0 @@
...
-import org.bukkit.event.Listener;
-
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
-
-/**
- * Class for accessing and managing linked accounts.
- */
-public abstract class AccountLinkManager implements Listener {
```
Vankka, line 34: Deprecating this is probably worth it (can even be converted back to regular class) - a lot of people use the linking system of DiscordSRV via the API

src/main/java/github/scarsz/discordsrv/linking/impl/system/sql/SqlAccountSystem.java
```diff
@@ -0,0 +1,420 @@
...
+            try {
+                connection.close();
+            } catch (SQLException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    @Override
+    @SneakyThrows
```
Vankka, line 159: A lot of SneakyThrows on user facing methods (SQL- and RedisAccountSystems), I don't think it's a great idea to throw a SQLException to the end user here. Wrapping to a exception (checked or unchecked) that says that "DiscordSRV failed to lookup the linked user" or something along those lines would be nice.

src/main/java/github/scarsz/discordsrv/linking/AccountSystem.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.UUID;
+
+public interface AccountSystem extends DiscordAccountStore, MinecraftAccountStore, CodeStore {
+
+    @NotNull AccountLinkResult process(String code, String discordId);
+    int getLinkCount();
+
+    default void link(UUID playerUuid, String discordId) {
```
Vankka, line 37: There doesn't seem to be a mechanism for this failing/throwing anything even if the backend is mcauth? (would be useful as this is user facing)

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/linking/impl/system/BaseAccountSystem.java
```diff
@@ -39,57 +37,76 @@
...
+
+
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
Scarsz: Leftover whitespace from my little work area when I was transferring everything over from the old implementations to the new implementations

src/main/java/github/scarsz/discordsrv/linking/impl/system/sql/MySQLAccountSystem.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Creates a MySQL SQL-backed account system utilizing a file-based H2 database
+     * @param host the host of the MySQL server
+     * @param port the port of the MySQL server
+     * @param database the database to be accessed
+     * @param username the username to authenticate with
+     * @param password the password for the given database user
+     */
+    public MySQLAccountSystem(String host, int port, String database, String username, String password) throws SQLException {
+        Driver mysqlDriver = new Driver();
```
Vankka, line 51: https://github.com/DiscordSRV/DiscordSRV/issues/1089

Scarsz: I assumed this was the correct way to do it now, feel free to change it

Vankka: This is using the regular class that's not in the `cj` package, but that may not always be provided since bukkit normally provides a version that doesn't have the `cj` package at all. This should be able to use the newer class when it's available

src/main/java/github/scarsz/discordsrv/linking/AccountSystem.java
Scarsz: MCAuth isn't a valid backend currently

PR #1156, opened 2021-06-11
Hook for Herochat Pro

PlayingWithPi:
With HC Pro being private and all, some of their imports were re-worked and didn't link back to the correct functions...

This is the updated import calls.
Created it as a separate Hook to Herochat.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/hooks/chat/HerochatProHook.java
+++ b/src/main/java/github/scarsz/discordsrv/hooks/chat/HerochatProHook.java
@@ -1,0 +1,102 @@
+/*-
+ * LICENSE
+ * DiscordSRV
+ * -------------
+ * Copyright (C) 2016 - 2021 Austin "Scarsz" Shapiro
+ * -------------
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/gpl-3.0.html>.
+ * END
+ */
+
+package github.scarsz.discordsrv.hooks.chat;
+
+/* Herochat Pro Imports */
+import com.herochat.api.Channel;
+import com.herochat.api.event.ChannelChatEvent;
+import com.herochat.api.Chatter;
+import com.herochat.HeroChat;
+
+/* DiscordSRV Imports */
+import github.scarsz.discordsrv.DiscordSRV;
+import github.scarsz.discordsrv.util.LangUtil;
+import github.scarsz.discordsrv.util.MessageUtil;
+import github.scarsz.discordsrv.util.PlayerUtil;
+import github.scarsz.discordsrv.util.PluginUtil;
+
+/* Other Imports */
+import net.kyori.adventure.text.Component;
+import org.apache.commons.lang3.StringUtils;
+import org.bukkit.event.EventHandler;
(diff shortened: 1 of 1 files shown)


src/main/java/github/scarsz/discordsrv/hooks/chat/HerochatProHook.java
```diff
@@ -0,0 +1,102 @@
...
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/gpl-3.0.html>.
+ * END
+ */
+
+package github.scarsz.discordsrv.hooks.chat;
+
+/* Herochat Pro Imports */
```
Vankka, line 25: No comments on imports

src/main/java/github/scarsz/discordsrv/hooks/chat/HerochatProHook.java
```diff
@@ -0,0 +1,102 @@
...
+                    return channel;
+                }
+            }
+            DiscordSRV.debug("No matching Herochat channels for name \"" + name + "\"");
+        } else {
+            DiscordSRV.debug("Herochat's channel manager returned no registered channels");
+        }
+        return null;
+    }
+
```
Vankka, line 96: isEnabled should be overridden, and it should check if the HerochatPro classes are available - if they are returning true otherwise false.
The same should be done for the regular Herochat hook, except checking if non-pro classes are available

Vankka (changes requested):

PlayingWithPi (commented):
Sounds good to me (👍🏼

src/main/java/github/scarsz/discordsrv/hooks/chat/HerochatProHook.java
Scarsz: Vankka wasn't saying that the classes should be merged, just that the ChatHook#isEnabled method should be overridden to be more context sensitive for HeroChatPro

PlayingWithPi: Most of the Code above is just written out using the HerochatHook.java from DiscordSRV
If there’s any better implementation, or changes that you’d like to make to the above, feel free to do so. I was only updating the imports for the v6 plugin in the file above (: @Scarsz

Androkai:
Also, since the API has changed, a publicly available jar of the HeroChat v6 API is needed to compile all of this. We need to add the jar to a Maven repo and reference it in pom.xml.

PlayingWithPi:
Let me talk to the Owner of the plugin and see what I can get sorted for you guys :)

PlayingWithPi:
@Androkai is it just the API you need access to? Or the full version of the .jar file?

Just wanting to confirm confirm it beforehand

Vankka:
Just the API, in order to compile

granny:
@PlayingWithPi any update?

PlayingWithPi:
> @PlayingWithPi any update?

Hey, just gotten a reply from them, they’ve given me the whole jar file, which is the entire plugin …
Posting it here would make people have access to the current production version of it.

Is there any way I could drop it in someone’s DM’s, to have it shared between the team only?

granny:
> > @PlayingWithPi any update?
> 
> Hey, just gotten a reply from them, they’ve given me the whole jar file, which is the entire plugin …
> Posting it here would make people have access to the current production version of it.
> 
> Is there any way I could drop it in someone’s DM’s, to have it shared between the team only?

Send it to me on discord; `granny#0429`

Androkai:
It seems to me that a brief explanation is needed here: In order for us to include the hook on our site, we need a publicly available jar that contains at least the API of the plugin. Since we add this to our Maven repo, anyone with any idea how Maven works can access and download this file. Since you're not sending us the file publicly, I don't think this is in the HeroChat developer's best interest.

For this reason, usually developers of chat plugins that are not freely available add DiscordSRV integration to their site using our API or publish their API separately.

PlayingWithPi:
I’ll pass on the message and see what they say

PlayingWithPi:
http://nexus.hc.to/content/repositories/pub_snapshots/com/herochat/Herochat/5.7.0-SNAPSHOT/

The link to the public snapshot.

PR #1188, opened 2021-07-07
Fix the {BOT} placeholder displaying an incorrect tag

Dinty1:
If the bot had a nickname in the server, the tag would have shown up as nickname#discrim

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/modules/requirelink/RequireLinkModule.java
+++ b/src/main/java/github/scarsz/discordsrv/modules/requirelink/RequireLinkModule.java
@@ -29,6 +29,7 @@ import github.scarsz.discordsrv.util.MessageUtil;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.User;
 import org.apache.commons.lang3.StringUtils;
 import org.bukkit.Bukkit;
 import org.bukkit.OfflinePlayer;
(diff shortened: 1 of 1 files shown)


Scarsz:
Why is this an issue? People looking for the bot in the user list aren't going to be looking for the username of the bot

MaximMaximS:
But if you want to display a nickname then don't display the discriminator, because Nickname#Discrim is nonsense.

Scarsz:
> But if you want to display a nickname then don't display the discriminator, because Nickname#Discrim is nonsense.
Good luck finding the correct bot in a community that have named all the server bots the same thing.

> And personally, I think Username#Discrim is better.
You can't find that in the user list, literally defeating the entire purpose of mentioning the name of the bot.

MaximMaximS:
> Good luck finding the correct bot in a community that has named all the server bots the same thing.

Good luck trying to find Nickname#Discrim in Ctrl + K, and if someone names all bots the same, then you still have to open each bot to check the discriminator, so using Username would have the same effect.

> You can't find that in the user list, literally defeating the entire purpose of mentioning the name of the bot.

Yes, you can't, but if you show Nickname#Discrim and someone smart will put it in Ctrl + K (because it is faster than looking for the bot yourself), they will find nothing.

Then why just not show Nickname (Username#Discrim)?

Scarsz:
> ... Ctrl + K ...

You're assuming the overwhelming majority a) know what Ctrl + K even is and b) wants to use a shortcut to find the bot. They don't.

> Then why just not show Nickname (Username#Discrim)?

There's certainly an argument to be made for that. I never said that wasn't a possibility. That said though, this is a user-customizable message and in all the years DiscordSRV has incorporated account linking, this is the very first time someone has even suggested that the current format has flaws.

MaximMaximS:
>You're assuming the overwhelming majority a) know what Ctrl + K even is and b) wants to use a shortcut to find the bot. They don't.

You're assuming that **EVERYONE** will be trying to find the bot in the user list because they won't be able to find it anywhere else (Ctrl + K) due to misleading information, which is applicable in only one situation. I think Nickname (Username#Discrim) would be the best solution applicable in all cases.

granny:
This is getting ridiculous. The message format is completely customizable. You don't even need to use the placeholder. If it's really that important for people, they can type the user tag manually. Thanks.

PR #1199, opened 2021-07-19
Make AvatarURL use mc-heads.net

James-Bennett-295:
Using mc-heads.net means the {texture} placeholder can be used to get the head so Discord caching is not an issue. Another benefit of this is people using offline mode servers would not have to change the URL.


Dinty1 (changes requested):
You should probably remove the warning about offline mode avatars now (and the bit where it changes what actually is used).

James-Bennett-295:
oops I edited the file from master branch, let me sort that sorry

Vankka (changes requested):
I don't know that I'm a fan of changing the default service, because we don't know how reliable the service is. Additionally texture is __not__ available on every server

PR #1202, opened 2021-07-21
Update README.

RezzedUp:
- Fix code size badge (before it was simply displaying the code size of the [badges/shields](https://github.com/badges/shields) repo)
- Fix link alt text
- Add bStats graph (if this isn't desired I can simply remove the commit before merging)


Scarsz:
bStats should probably be on top of the thank-yous

PR #1211, opened 2021-08-01
Update de.yml

KagetaneHiruko:
Changed {EINLADUNG} to {INVITE} so it works again and fixed some borked translation


Androkai:
Looks good, thank you.

PR #1229, opened 2021-08-17
Allow for newlines (\n) in /discord broadcast

ANutley:
Title is self explanatory.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/commands/CommandBroadcast.java
+++ b/src/main/java/github/scarsz/discordsrv/commands/CommandBroadcast.java
@@ -45,7 +45,7 @@ public class CommandBroadcast {
         TextChannel target = null;
 
         if (args.length > 0 && args[0].startsWith("#")) {
-            String raw = args[0].replace("#", "").replace("\\n", "\n");
+            String raw = args[0].replace("#", "");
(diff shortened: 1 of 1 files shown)


Vankka:
This is a breaking change to the way the broadcast command works, some users might have `\n` in the content of broadcast messages (while not wanting them to be converted to newlines)

Vankka (changes requested):
See above

Dinty1:
A config option set to false by default could solve that problem

Vankka (approved):
After deliberation, use cases where `\n` is wanted plain are unlikely enough that, having this minor breakage is fine.

src/main/java/github/scarsz/discordsrv/commands/CommandBroadcast.java
```diff
@@ -45,7 +45,7 @@ public static void execute(CommandSender sender, String[] args) {
         TextChannel target = null;
 
         if (args.length > 0 && args[0].startsWith("#")) {
-            String raw = args[0].replace("#", "");
+            String raw = args[0].replace("#", "").replace("\\n", "\n");
```
Vankka, line 48: This is only used to get the channel, thus it isn't neccasery to replace newlines here

ANutley: Changed

PR #1248, opened 2021-09-11
Update plugin.yml to include the /discord debugger command

ANutley:


src/main/resources/plugin.yml
```diff
@@ -35,6 +35,7 @@ commands:
            /discord reload - Reload the plugin
            /discord resync - Resynchronizes all groups & roles
            /discord language - Changes the language of DiscordSRV to whatever is specified
+           /discord debugger - Enables debug mode temporarily for either all or a certain debug category
```
Dinty1, line 38: Should probably say something similar to https://github.com/DiscordSRV/DiscordSRV/blob/master/src/main/java/github/scarsz/discordsrv/commands/CommandDebugger.java#L35

Dinty1 (changes requested):

src/main/resources/plugin.yml
ANutley: Updated

Dinty1 (approved):

PR #1252, opened 2021-09-13
More appropriate watchdog timeout default message.

ipaqmaster:
Was pretty worried when I saw this @me this afternoon while away from home thinking some sort of temperature warning threshold were met, or any kind of alarming scenario. Instead it was just a plugin on the Survival instance acting up and making the server's gameticks take a while, activating this watchdog message.

Changed to a more appropriate default message.

### DIFF
--- a/src/main/resources/messages/en.yml
+++ b/src/main/resources/messages/en.yml
@@ -397,7 +397,7 @@ DiscordChatChannelServerShutdownMessage: ":octagonal_sign: **Server has stopped*
 #                        you can @mention the owner of the server by using "%guildowner%"
 #                        you can put the date and time of the crash in the message by using %date%
 #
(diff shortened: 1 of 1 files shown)


ipaqmaster:
See [this](https://www.reddit.com/r/servers/comments/bmmhki/im_scared_i_run_my_server_in_my_house_and_its/) reddit post for a familiar experience to what I had this afternoon.

Scarsz (changes requested):
The new message isn't the best in my opinion. If it's going to be changed, it should be changed to be clear and concise. Maybe something like `the server hasn't ticked in 30 seconds`

The message needs to be changed for the other languages as well

ipaqmaster:
I agree, this was a pretty quick change I made in my own copy as well however your suggestion would be a significantly better message already.

Dinty1:
May also be worth incorporating #1259 into this PR

PR #1260, opened 2021-09-18
Cancelling double leave message - paper bug

felix642:
There's currently a bug in paper where the leave event would sometimes fire twice resulting in the leave message appearing twice in discord.
Screenshot from 2021-09-17 20-59-55
This pull request adds a condition to the leave event to ignore the second message.


src/main/java/github/scarsz/discordsrv/listeners/PlayerJoinLeaveListener.java
```diff
@@ -106,6 +106,11 @@ public void onPlayerJoin(PlayerJoinEvent event) {
     @EventHandler(priority = EventPriority.LOWEST) //priority needs to be different to MONITOR to avoid problems with permissions check when PEX is used, it is at lowest so that it executes before VanishNoPacket's player leave listener and is able to see whether the player is vanished
     public void PlayerQuitEvent(PlayerQuitEvent event) {
         final Player player = event.getPlayer();
+        if(!player.isOnline()) {
```
Dinty1, line 109: Needs a space between the `if` and the `(`

Dinty1 (changes requested):

Vankka:
Shouldn't the paper bug be patched instead of making a workaround here..?

granny:
This isn't a bug fix. This is a bandaid fix for a problem that we don't have.

PR #1294, opened 2021-10-26
Implement degraded client notice & legacifying of messages for LunarClient

Vankka:
For users stumbling upon this PR,

Due to LunarClient hiding messages containing click to copy functionality (this also happens for vanilla commands, eg. the `/seed` command) and not showing any intention to fix this even after being notified of the issue (multiple times over 5+ months), we have decided to add a notification for users using LunarClient letting them know that some commands may not work as expected, as this functionality should work as it does on the vanilla client instead of **completely hiding the message**.

[Image](https://i.imgur.com/uyU5RyV.png) of their support *manually* closing a ticket without comment regarding this issue. The ticket was closed before the last message. The last message was sent in hopes of response to no avail.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -52,6 +52,7 @@ import github.scarsz.discordsrv.objects.log4j.JdaFilter;
 import github.scarsz.discordsrv.objects.managers.AccountLinkManager;
 import github.scarsz.discordsrv.objects.managers.CommandManager;
 import github.scarsz.discordsrv.objects.managers.GroupSynchronizationManager;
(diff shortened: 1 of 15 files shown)


Scarsz (approved):

Vankka (commented):
Confirmed still broken on LunarClient 1.17.1-63e9251. Tested that channel register, plugin message and client brand will trigger the notice.

Badbird-5907:
> Due to LunarClient hiding messages containing click to copy functionality (this also happens for vanilla commands, eg. the /seed command)

This seems to be fixed now? (Please correct me if I'm wrong)
image

Vankka:
Will check again, soon

Vankka:
Confirmed that LunarClient fixed it, took them long enough. Next version has the notice & click/hover component removal taken out

Vankka:
The latest release still has it, it will be gone in the next version

Issue #861, opened 2020-09-11
Add support to block certain roles from having their messages processed and send from Discord to MC


acollierr17:
I am proposing to block specific roles from speaking in "bridge" channels defined in the DiscordSRV config. I've noticed that specific [users based on their IDs](https://github.com/DiscordSRV/DiscordSRV/wiki/config#discordchatchannelblockedids) can be blocked, but not roles. An example could be a user with a linked Minecraft account but has been muted in the Discord (user is added to the `Muted` role). They can still send messages just fine in the "bridge" channel, which can pose an issue (this is because of how Discord's permissions system is [set up](https://discordjs.guide/popular-topics/permissions-extended.html#discord-s-permission-system)). I haven't found anything in the Wiki via the search term `role`.

Scarsz:
This should be handled instead by denying the role's write permission to the channel

acollierr17:
> This should be handled instead by denying the role's write permission to the channel

I stated this isn't possible due to Discord's permissions system. For example, if the `@everyone` role globally has `SEND_MESSAGES`, but you have the linked role configured as `Verified` and its permission in the bridge channel, the `SEND_MESSAGES` is explicitly set to true, the user can still send messages, even when the `Muted` role's permission is set to false. Please check the second link in my issue instead of closing it immediately.

Scarsz:
> I stated this isn't possible due to Discord's permissions system.

You didn't explain your situation thoroughly enough for it to be apparent *why* this is a problem with Discord's permission system.

> Please check the second link in my issue instead of closing it immediately.

I'm aware of how Discord permissions work.

After you've given an explanation, I agree that this could be something that's implemented.

AlaRubra123:
> > This should be handled instead by denying the role's write permission to the channel
> 
> I stated this isn't possible due to Discord's permissions system. For example, if the `@everyone` role globally has `SEND_MESSAGES`, but you have the linked role configured as `Verified` and its role added in the bridge channel ith the `SEND_MESSAGES` permission explicitly set to true, the user can still send messages, even when the `Muted` role's permission is set to false. Please check the second link in my issue instead of closing it immediately.
> 
> Edit: Fix some typos and grammatical errors

Honestly, that just sounds like your permissions need to be edited. Like, for a verification role, make `@everyone` have `SEND_MESSAGES` off globally, then have `Verified` have it on globally, and leave channel permissions for both roles neutral.

Then, your muted role will work correctly.

[...]

PR #1297, opened 2021-10-28
Block specified roles from chat channel

vaporrrr:
Implements #861


Vankka (approved):

Scarsz (changes requested):
I feel like this should probably inform the user that their message wasn't being sent to Minecraft. Whether that be a ❌ reaction or simply deleting the message is up for debate, but I'm not a fan of silently ignoring messages.

Dinty1 (approved):
LGTM

Vankka (approved):

PR #1302, opened 2021-11-11
Allow you to change the online status of the bot (online, idle, invisible and do not disturb)

ANutley:


src/main/java/github/scarsz/discordsrv/objects/threads/PresenceUpdater.java
```diff
@@ -77,27 +80,44 @@ public void run() {
                     status = PlaceholderUtil.replacePlaceholders(status);
                     status = MessageUtil.strip(status); // remove color codes
                 }
-                boolean same = Objects.equals(lastStatus, status);
+
+                String onlineStatusString = DiscordSRV.config().getString("DiscordOnlineStatus").toUpperCase(Locale.ROOT).trim();
+                OnlineStatus onlineStatus = OnlineStatus.ONLINE;
```
Scarsz, line 85: Assignment should be in a `default` case

src/main/java/github/scarsz/discordsrv/objects/threads/PresenceUpdater.java
```diff
@@ -77,27 +80,44 @@ public void run() {
...
+
+                switch (onlineStatusString) {
```
Scarsz, line 87: No reason for this switch to exist when the configured values are identical to the enum values, Enum#valueOf(name) would be preferred

Having the configured values be the same isn't good either though. Should be something like `dnd` or at least get rid of the `_`'s with something like `do not disturb`

src/main/java/github/scarsz/discordsrv/objects/threads/PresenceUpdater.java
```diff
@@ -29,18 +29,21 @@
...
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 public class PresenceUpdater extends Thread {
 
     private int lastStatusIndex = 0;
     private String lastStatus = null;
+    private String lastOnlineStatus = null;
```
Scarsz, line 46: Why is this a String? The lastStatus var is a String because it's a message. This is an enum, it should just be stored as the enum.

src/main/resources/config/en.yml
```diff
@@ -124,8 +124,10 @@ UseModernPaperChatEvent: false
 # PlaceholderAPI placeholders are supported
 #
 # DiscordGameStatus: Text to display. Can be a single value e.g. "Minecraft", or multiple values e.g. ["Minecraft", "yourip.changeme.com"]
+# DiscordOnlineStatus: Activity status to display. This must be one of the following: ONLINE, DO_NOT_DISTURB, IDLE or INVISIBLE
 # StatusUpdateRateInMinutes: How frequently to cycle the status
 DiscordGameStatus: ["Minecraft"]
+DiscordOnlineStatus: "ONLINE"
```
Scarsz, line 130: Quotes not necessary for simple string values

Scarsz (changes requested):

src/main/java/github/scarsz/discordsrv/objects/threads/PresenceUpdater.java
```diff
@@ -77,27 +80,34 @@ public void run() {
...
+                if (onlineStatus == OnlineStatus.UNKNOWN) onlineStatus = OnlineStatus.ONLINE;
+
+                boolean same = Objects.equals(lastStatus, status) && Objects.equals(onlineStatus, lastOnlineStatus);
                 lastStatus = status;
+                lastOnlineStatus = onlineStatus;
 
                 if (!same) {
                     if (StringUtils.isNotBlank(status)) {
-                        DiscordSRV.debug(Debug.PRESENCE, "Setting presence to \"" + status + "\"");
+                        DiscordSRV.debug(Debug.PRESENCE, "Setting presence to \"" + status + "\"" + (onlineStatus != OnlineStatus.ONLINE ? " and online status to \"" + onlineStatusString + "\"" : ""));
```
Vankka, line 95: Hiding that it's changing the status to online is pointless

Vankka (approved):

PR #1303, opened 2021-11-12
Unix Timestamp Placeholders

### LINKED ISSUE #1271: UNIX Timestamps in Placeholders
### Why?

This could be used for Discord's date feature to more easily get times in your timezone and relative to you, for example, last updated time could be relative to the current time. This could also lower the amount of updates something like a channel topic would get if you dynamically update it with the uptime.

### What and How?

Maybe a placeholder like
`%unixtime%`

There could be UNIX timestamps for other times too, such as when the server started, for example:
`%unixstarttime%`

Or maybe when the server shuts down, it could be used for relative shutdown time, like:
`%unixshutdowntime%`

### Are there alternatives?

No visible alternative for current time, the closest you can get for uptime would be the `%uptimemins%` and `%uptimehours%` placeholders, though a UNIX timestamp version would update faster. Downtime has no alternative because the server could not update a channel topic if it was offline.

### Checks

### Anything else

With something like channel topics, it could be used like this:

"Server went offline [10 minutes ago]"
In the config it could look like:
`ChannelTopicUpdaterChatChannelTopicAtServerShutdownFormat: "Server went offline <t:%unixshutdowntime%:R>"`

"Server started [5 hours ago]"
Config:
`ChannelTopicUpdaterChatChannelTopicFormat: "Server started <t:%unixstarttime%:R>"`

Cale111:
This would close #1271


Vankka (changes requested):
Please use spaces instead of tabs for indentation in the Java files.

Please also add a extra space between the placeholder and it's description in the messages files, and align all of the other placeholder descriptions to match the indentation, eg.

```diff
...
- # %timestamp%:     current time in seconds since UNIX epoch
- # %starttimestamp%:start time in seconds since UNIX epoch
- # %totalplayers%:  total amount of players to ever join the main world
+ # %timestamp%:      current time in seconds since UNIX epoch
+ # %starttimestamp%: start time in seconds since UNIX epoch
+ # %totalplayers%:   total amount of players to ever join the main world
...
```

src/main/resources/messages/en.yml
```diff
@@ -124,6 +124,8 @@ DynmapDiscordFormat: "[Dynmap] %name% » %message%"
 # Available placeholders:
 # %datetime%:  current date & time
 #               example: Sun Jan 1 15:30:45 PDT 2017
+# %timestamp%: current time in seconds since unix epoch
+#               example: 1483313445
```
Vankka, line 128: This is missing in files besides English

Vankka (approved):

Dinty1 (approved):

Issue #1135, opened 2021-05-26
Add config option for the mention sound


arts7493:
add config option plz
mention sound in minecraft when someone mentioned

dani626:
https://www.spigotmc.org/resources/interactivechat-show-items-inventory-in-chat-custom-chat-keywords-bungee-velocity-support.75870/

Dinty1:
> https://www.spigotmc.org/resources/interactivechat-show-items-inventory-in-chat-custom-chat-keywords-bungee-velocity-support.75870/

...Not really sure what you mean. The mention sound is a part of DiscordSRV.

Scarsz:
Implemented

PR #1307, opened 2021-11-24
Allows the mention sound in Minecraft to be disabled

ANutley:
Addresses #1135


src/main/java/github/scarsz/discordsrv/util/PlayerUtil.java
```diff
@@ -116,6 +116,8 @@ public static void notifyPlayersOfMentions(Predicate<? super Player> predicate,
                         })
                         .collect(Collectors.toList());
 
+        if (!DiscordSRV.config().getBoolean("MinecraftMentionSound")) return;
```
Scarsz, line 119: Should short circuit higher in the method, after the predicate null check preferably

Scarsz (changes requested):
`MinecraftMentionSound` being on it's own in the bottom of the file isn't very attractive imo, would be better in the `Other` category

PR #1309, opened 2021-11-25
Replace logging appender with JDA Appender library

Scarsz:
- Automatically updates existing messages until they are full
- Color codes different logging levels
- Customizable prefix/suffixes, logger padding amount


src/main/java/github/scarsz/discordsrv/listeners/DiscordConsoleListener.java
```diff
@@ -71,23 +71,14 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
...
-        boolean allowed = DiscordConsoleChannelBlacklistActsAsWhitelist == DiscordConsoleChannelBlacklistedCommands.contains(requestedCommand);
-        // return if command not allowed
-        if (!allowed) return;
+        boolean isWhitelist = DiscordSRV.config().getBoolean("DiscordConsoleChannelBlacklistActsAsWhitelist");
+        List<String> blacklistedCommands = DiscordSRV.config().getStringList("DiscordConsoleChannelBlacklistedCommands");
+
+        for (int i = 0; i < blacklistedCommands.size(); i++) blacklistedCommands.set(i, blacklistedCommands.get(i).toLowerCase());
+
+        String requestedCommand = event.getMessage().getContentRaw().trim().split(" ")[0].toLowerCase();
+        requestedCommand = requestedCommand.substring(requestedCommand.lastIndexOf(":")) + 1;
```
Vankka, line 80:
```suggestion
        requestedCommand = requestedCommand.substring(requestedCommand.lastIndexOf(":") + 1);
```

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -1418,9 +1436,6 @@ public void onDisable() {
                 // kill channel topic updater
                 if (channelTopicUpdater != null) channelTopicUpdater.interrupt();
 
-                // kill console message queue worker
-                if (consoleMessageQueueWorker != null) consoleMessageQueueWorker.interrupt();
-
```
Vankka, line 1423: The appender should be shutdown at disable

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
Scarsz: [It is](https://github.com/DiscordSRV/DiscordSRV/pull/1309/files#diff-cafc7a5fe69034c557bfa518b2df149d2d01fbaa7064a31f5aeb8019ae95471eR1458-R1459)

Vankka: Link goes to nowhere, but I see

Vankka (approved):

Scarsz (commented):
Going to add a short date/timestamp to the default prefixer, likely just `mm/dd/yy hh:mm:ss` to keep it short & concise

Vankka (commented):
Waiting on fix to prefixes being the same for all rows in a message

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -996,19 +996,37 @@ public void connectFailed(URI arg0, SocketAddress arg1, IOException arg2) {}
...
+                    String timestamp = TimeUtil.consoleTimeStamp(item.getTimestamp());
+                    return PlaceholderUtil.replacePlaceholdersToDiscord(config().getString(key))
+                            .replace("{date}", timestamp)
+                            .replace("{datetime}", timestamp)
+                            .replace("{name}", StringUtils.isNotBlank(name) ? " " + name : "")
+                            .replace("{level}", config.padLevelName(item.getLevel().name()));
+                };
+                config.setPrefixer(item -> placeholders.apply("DiscordConsoleChannelPrefix", item));
+                config.setSuffixer(item -> placeholders.apply("DiscordConsoleChannelSuffix", item));
+            }).attachLog4jLogging().schedule();
```
Vankka, line 1029: Scheduling updates every 1 second will lead to a infinite rate limit loop if Discord ever has even the tiniest outage (already observed with Melanie)

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
Scarsz: What would you suggest? 1.5 seconds?

Vankka: I'm ok with that, atleast that will clear up the rate limit queue after a while. 

Ideally avoiding building the "queue" up to begin with would be nice, but probably isn't very doable in practice

Issue #1318, opened 2021-12-07
Add/unblock webhooks


Recontaminated:
### Why?

Currently, webhooks sent in a chat channel won't forward to Minecraft, a lot of simple bots that just need to post text use webhooks and it sucks that the only way around this is to make a bot.

### What and How?

Webhooks allow you to specify the username they show as, so in Minecraft, it would look exactly like any other user. Theres currently "DiscordChatChannelBlockBots: false" but even when set to false, it still blocks webhook messages.

### Are there alternatives?

Yeah, it could be placeholder text or anything that just gets webhooks inside of minectaft.

### Checks

- [X] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

### Anything else

Asked if there was a workaround in discord support ticket #6376

PR #1334, opened 2021-12-13
Allow webhooks to be used in Discord -> MC

ANutley:
Implements #1318

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -100,15 +100,15 @@ public class DiscordChatListener extends ListenerAdapter {
         if (!DiscordSRV.config().getBoolean("DiscordChatChannelDiscordToMinecraft")) return;
 
         // enforce required account linking
-        if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount")) {
-            if (!event.getAuthor().isBot() && DiscordSRV.getPlugin().getAccountLinkManager() == null) {
+        if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount") && !event.getAuthor().isBot()) {
+            if (DiscordSRV.getPlugin().getAccountLinkManager() == null) {
                 event.getAuthor().openPrivateChannel().queue(privateChannel -> privateChannel.sendMessage(LangUtil.Message.FAILED_TO_CHECK_LINKED_ACCOUNT.toString()).queue());
                 DiscordUtil.deleteMessage(event.getMessage());
                 return;
             }
 
             boolean hasLinkedAccount = DiscordSRV.getPlugin().getAccountLinkManager().getUuid(event.getAuthor().getId()) != null;
-            if (!hasLinkedAccount && !event.getAuthor().isBot()) {
+            if (!hasLinkedAccount) {
                 LangUtil.Message formatOption = LangUtil.Message.LINKED_ACCOUNT_REQUIRED;
                 String format = formatOption.toString();
                 if (format != null && !format.isEmpty()) {
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -33,7 +33,6 @@ import github.scarsz.discordsrv.objects.SingleCommandSender;
 import github.scarsz.discordsrv.util.*;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.Role;
-import net.dv8tion.jda.api.entities.Webhook;
 import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.kyori.adventure.text.Component;
@@ -70,11 +69,9 @@ public class DiscordChatListener extends ListenerAdapter {
                 return;
             }
 
-            Webhook webhook = event.getChannel().retrieveWebhooks().complete().stream()
-                    .filter(hook -> hook.getName().equals("DiscordSRV " + event.getChannel().getId()))
-                    .findFirst().orElse(null);
+           String webhookId = WebhookUtil.getWebhookUrlToUseForChannel(event.getChannel()).split("/")[5];
 
-            if (webhook != null && webhook.getId().equals(event.getAuthor().getId())) return;
+            if (webhookId.equals(event.getAuthor().getId())) return;
         }
 
         // canned responses
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -69,8 +69,8 @@ public class DiscordChatListener extends ListenerAdapter {
                 return;
             }
 
-           String webhook = WebhookUtil.getWebhookUrlFromCache(event.getChannel());
-
+            // Prevent our own webhook from being picked up
+            String webhook = WebhookUtil.getWebhookUrlFromCache(event.getChannel());
             if (webhook != null && webhook.split("/")[6].equals(event.getAuthor().getId())) return;
         }
 
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -33,8 +33,11 @@ import github.scarsz.discordsrv.objects.SingleCommandSender;
 import github.scarsz.discordsrv.util.*;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.Webhook;
 import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
+import net.dv8tion.jda.api.exceptions.ErrorHandler;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
+import net.dv8tion.jda.api.requests.ErrorResponse;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.TextReplacementConfig;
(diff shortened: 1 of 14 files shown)


src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -59,9 +62,16 @@
     @Override
     public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
         // if message is from null author or self do not process
-        if (event.getMember() == null || DiscordUtil.getJda() == null || event.getAuthor().equals(DiscordUtil.getJda().getSelfUser()))
+        if ((event.getMember() == null && !event.isWebhookMessage()) || DiscordUtil.getJda() == null || event.getAuthor().equals(DiscordUtil.getJda().getSelfUser()))
             return;
 
+        Webhook webhook = event.getChannel().retrieveWebhooks().complete().stream()
+                .filter(hook -> hook.getName().equals("DiscordSRV " + event.getChannel().getId()))
+                .findFirst().orElse(null);
```
Vankka, line 70: This request should be avoided by caching the ID instead of **getting all webhooks every time there is a Discord message**

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -59,9 +62,16 @@
...
+
+        if (webhook != null)
+            if (webhook.getId().equals(event.getAuthor().getId())) return;
```
Vankka, line 73: Use `if (condition1 && condition2)` rather than two seperate `if`s in places like this. This check is not necessary if `DiscordChatChannelBlockWebhooks` is enabled (effectively eliminating the need to know the webhook id)

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -90,7 +100,7 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
         // enforce required account linking
         if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount")) {
             if (DiscordSRV.getPlugin().getAccountLinkManager() == null) {
-                event.getAuthor().openPrivateChannel().queue(privateChannel -> privateChannel.sendMessage(LangUtil.Message.FAILED_TO_CHECK_LINKED_ACCOUNT.toString()).queue());
+                event.getAuthor().openPrivateChannel().queue(privateChannel -> privateChannel.sendMessage(LangUtil.Message.FAILED_TO_CHECK_LINKED_ACCOUNT.toString()).queue(null, new ErrorHandler().ignore(ErrorResponse.CANNOT_SEND_TO_USER)));
```
Vankka, line 103: Sure ignoring this exception is fine, but that is not a good solution to avoiding error responses when sending **invalid requests** (event.getAuthor() is a fake user when the message is from a webhook)

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -125,7 +135,7 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
                         output = format.replace(placeholder, msg);
                     }
                     event.getAuthor().openPrivateChannel().queue(privateChannel ->
-                            privateChannel.sendMessage(output).queue());
+                            privateChannel.sendMessage(output).queue(null, new ErrorHandler().ignore(ErrorResponse.CANNOT_SEND_TO_USER)));
```
Vankka, line 138: Sure ignoring this exception is fine, but that is not a good solution to avoiding error responses when sending **invalid requests** (event.getAuthor() is a fake user when the message is from a webhook)

src/main/resources/config/en.yml
```diff
@@ -170,6 +171,7 @@ DiscordChatChannelRolesAllowedToUseColorCodesInChat: ["Developer", "Owner", "Adm
 DiscordChatChannelBroadcastDiscordMessagesToConsole: true
 DiscordChatChannelRequireLinkedAccount: false
 DiscordChatChannelBlockBots: false
+DiscordChatChannelBlockWebhooks: false
```
Vankka, line 174: Set this to `true` by default

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
ANutley: Without a reliable way to update the cache of the current webhooks in a channel, I instead made this part of the code only run if it is a webhook message and moved the `BlockWebhooks` check above this part to reduce the amount of times the webhooks are retrieved.

Vankka: This still needs to be resolved, technically the webhook id is already cached, since the webhook url is cached and the id is a part of it. You could even add caching for the webhook id along side the webhook url caching that's present right now

ANutley: I've used the WebhookUtil.getWebhookUrlToUseForChannel() method instead to hopefully resolve this.

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -59,9 +59,21 @@
...
             return;
 
+        // block webhooks
+        if (event.isWebhookMessage()) {
+            if (DiscordSRV.config().getBoolean("DiscordChatChannelBlockWebhooks")) {
+                DiscordSRV.debug(Debug.DISCORD_TO_MINECRAFT, "Received Discord message from webhook" + event.getAuthor() + " but DiscordChatChannelBlockWebhooks is on");
+                return;
+            }
+
+           String webhookId = WebhookUtil.getWebhookUrlToUseForChannel(event.getChannel()).split("/")[5];
```
Vankka, line 72: This might cause a webhook to be created unwatendly (when using webhooks is disabled)

ANutley: Updated

Vankka: It should be checking before it even tries to figure out the webhooks, otherwise might run into a permission exception

NoxInmortus:
Can't wait to see this PR merged!

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -59,9 +59,21 @@
...
+        // block webhooks
+        if (event.isWebhookMessage()) {
+            if (DiscordSRV.config().getBoolean("DiscordChatChannelBlockWebhooks")) {
+                DiscordSRV.debug(Debug.DISCORD_TO_MINECRAFT, "Received Discord message from webhook" + event.getAuthor() + " but DiscordChatChannelBlockWebhooks is on");
+                return;
+            }
+
+           String webhook = WebhookUtil.getWebhookUrlFromCache(event.getChannel());
+
+            if (webhook != null && webhook.split("/")[6].equals(event.getAuthor().getId())) return;
```
Vankka, line 74:
```suggestion
            // Prevent our own webhook from being picked up
            String webhook = WebhookUtil.getWebhookUrlFromCache(event.getChannel());
            if (webhook != null && webhook.split("/")[6].equals(event.getAuthor().getId())) return;
```

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -90,7 +102,8 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
         // enforce required account linking
         if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount")) {
             if (DiscordSRV.getPlugin().getAccountLinkManager() == null) {
-                event.getAuthor().openPrivateChannel().queue(privateChannel -> privateChannel.sendMessage(LangUtil.Message.FAILED_TO_CHECK_LINKED_ACCOUNT.toString()).queue());
+                if (!event.isWebhookMessage())
+                    event.getAuthor().openPrivateChannel().queue(privateChannel -> privateChannel.sendMessage(LangUtil.Message.FAILED_TO_CHECK_LINKED_ACCOUNT.toString()).queue());
                 DiscordUtil.deleteMessage(event.getMessage());
                 return;
             }
```
Vankka, line 109: Not directly related to this PR but this check should be skipped for bots, since they don't need a linked account anyway

Vankka (commented):
- Right now if `DiscordChatChannelBlockBots` is enabled, `DiscordChatChannelBlockWebhooks` becomes useless, `isBot` is true for webhooks.

This is a breaking change for API users who are expecting there to be a Member object

ParadoxFlame:
Will this PR ever be merged? Without this, dislink kinda breaks, and afaik dislink is the only way to have the same server live chat in multiple channels until DiscordSRV v2 comes out

Vankka:
Up to ANutley, there's review comments that need to be resolved before this can be merged

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -89,7 +101,7 @@ public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
 
         // enforce required account linking
         if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount")) {
```
Vankka, line 103:
```diff
-        if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount")) {
+        if (DiscordSRV.config().getBoolean("DiscordChatChannelRequireLinkedAccount") && !event.getAuthor().isBot()) {
```
& remove all the isBot checks within this if

Vankka (commented):
One thing that I thought of is that API users might expect non-null members from the events being posted here, but there's no not null annotations there either way, and working around those *potential expectations* seems like too high of an effort - so no action required there. 

See inline comment below, otherwise looks good

Vankka (approved):

thewilloftheshadow (approved):

PR #1354, opened 2022-01-17
Set code combinations from 10^4 to 35^4

dylhack:


src/main/java/github/scarsz/discordsrv/objects/managers/link/AbstractAccountLinkManager.java
```diff
@@ -8,12 +8,12 @@
...
  * published by the Free Software Foundation, either version 3 of the
  * License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
```
dylhack, line 16: I hope this isn't bothersome.

morfie78 (approved):
+1

Scarsz (changes requested):
Not at all a fan of using potentially ambiguous characters (l, lowercase L vs I, uppercase I) nor lowercase characters in general

dylhack:
I hope this helps!

dylhack:
Any further thoughts on this?

src/main/java/github/scarsz/discordsrv/objects/managers/link/FileAccountLinkManager.java
```diff
@@ -145,8 +145,11 @@ public String process(String linkCode, String discordId) {
             }
         }
 
-        // strip the code to get rid of non-numeric characters
-        linkCode = linkCode.replaceAll("[^0-9]", "");
+        // strip the code to get rid of non-numeric and non-alphabetical characters
+        linkCode = linkCode.replaceAll("[^A-Z0-9]", "");
+        // remove case sensitivity
+        linkCode = linkCode.toUpperCase();
```
Vankka, line 151: Please move the `toUpperCase` call above the regex, otherwise lowercase characters will not be picked up

Vankka (changes requested):
Sorry about the delayed review, other than this one comment this looks fine

dylhack:
Excellent, thanks for the response! I will get on that as soon as I can.

Scarsz (changes requested):
Still not really a fan of the ambiguous characters left such as 0 and O

morfie78:
As most people copy/paste the generated code, I don't think ambiguity is a problem.

Scarsz:
> As most people copy/paste the generated code, I don't think ambiguity is a problem.

Not true, especially when the server requires linking to play; the link code is absolutely not copy/pasteable from the kick screen.

morfie78:
You have a point.  @dylhack add a regex to change ambiguous chars?

dylhack:
I must argue and say that O and 0 aren't very alike, especially with Minecraft's typography. I can understand "I" and "l", but zero and O? Can you list off any more ambiguous characters?

image

Scarsz (changes requested):
I'm still really not a fan of having letters in the linking codes. If letters are behind a disabled by default config option, that would be fine. 6 characters seems overkill as well still but that change doesn't concern me as much.

In addition,

- https://github.com/dylhack/DiscordSRV/blob/develop/src/main/java/github/scarsz/discordsrv/objects/managers/link/JdbcAccountLinkManager.java#L367 needs to contain the updated strip pattern

- The message in all languages at https://github.com/dylhack/DiscordSRV/blob/develop/src/main/resources/messages/en.yml#L429 needs to be updated to reflect the new code style, probably just say codes are alphanumeric

PR #1401, opened 2022-05-08
Update to JDA 5

nightpool:
Hi all! Recently I found myself wanting to set up Discord linking for my small friend-group server, but we wanted to set it up so that minecraft messages went into a specific minecraft thread, to reduce channel clutter. In theory, this should be very easy, since threads are very very similar to channels under the hood. In practice, it required a lot of refactoring, because when the bot was first written there was only one type of TextChannel. However, I'm not the type to shrink from a challenge, so I sat down, did the refactorings, and now everything seems to be working! Here's an example:

image

From a technical perspective, the main change here is refactoring from using `TextChannel` everywhere to using either `MessageChannel` (if we don't need any specific `Guild` features) or `GuildMessageChannel` (chosen mostly for convenience—since a fair amount of this plugin relies on the functionality of the enclosing guild, and supporting DMs and group messages wasn't a requirement for me, it was a much simpler refactor to enforce the use of GuildMessageChannel for the main bot channels). Conceptually, there's no real *hard* dependencies on a sync channel existing inside of a guild, but in practice it's pretty useful for e.g. debugging, emoji, etc.

When changing the topic of a channel, we check to see if it implements `BaseGuildMessageChannel` ("base" in the sense that threads can have them as a parent), since only "base" channels can have a topic. If the configured channel does not have a topic, but topic editing is enabled, we warn. Similarly, only "base" channels can have webhooks configured, but you can still send webhooks to a specific thread by 1) creating a webhook on the base channel for that thread, and then 2) adding a `thread_id` query parameter to the webhook with the thread's channel ID, so a little bit of extra work was necessary there as well.

This relies in part on a similar refactor I did for `JDAAppender` to support this, see the corresponding PR here: https://github.com/Scarsz/JDAAppender/pull/8

There are a few other breaking changes in the DiscordSRV library (most notably, a user's voice session returns an `AudioChannel`, which may or may not not have a category associated with it). 

Since this requires some breaking changes to the API, I've bumped the version to `2.0.0-SNAPSHOT`. Let me know if you think that's a good idea, or if you preferred I use a different version for this breaking change.

I've tested this on my own server and everything seems to be working fine, but we only use a small fraction of the bot's features, and the changes are very wide-ranging. Special attention should be paid to the voice channel / category code, since I haven't had the chance to test any of it, unfortunately. (We don't use the voice channel feature on our server)

Thank you for reading! And thank you for the bot, the refactors were pretty painless, all things considered.


Vankka:
The reason we haven't done this yet is due to it breaking pretty much every other plugin relying on DiscordSRV due to JDA entities being exposed extensively, if we're gonna go through with this we'll want to make sure plugins using DiscordSRV get a heads up first.

I apprechiate you taking the time to do this, from the first glance this looks like great work under the hood :)

nightpool:
That makes sense. However, probably the first step that's necessary for
people having a heads up is having a SNAPSHOT build they can try compiling
against, right?

Happy to make WIP/example PRs for one or two downstream plugins if you have
any suggestions on which ones are most important / most difficult to tackle.

Vankka:
> However, probably the first step that's necessary for people having a heads up is having a SNAPSHOT build they can try compiling against, right?

Yes, but that's a bit problematic as JDA is [making breaking changes](https://github.com/DV8FromTheWorld/JDA/pulls?q=is%3Apr+is%3Aopen+label%3A%22type%3A+breaking%22) as they are still in 'alpha' with version 5, so that might quickly become a major pain if every new version of DiscordSRV needs a update in plugins that depend on it because of breaking changes.

> Happy to make WIP/example PRs for one or two downstream plugins if you have any suggestions on which ones are most important / most difficult to tackle.

The biggest plugins depends on DiscordSRV's api that I can think of would probably be CMI and ChatControl(Red) but those are closed source. Some open source ones that come to mind are [Plan](https://github.com/plan-player-analytics/Extension-DiscordSRV) and [InteractiveChat's addon](https://github.com/LOOHP/InteractiveChat-DiscordSRV-Addon), there's also some others listed here: https://docs.discordsrv.com/#plugins-that-use-our-api - but I'd still hold off on those due to the point I made above.

> Since this requires some breaking changes to the API, I've bumped the version to 2.0.0-SNAPSHOT. Let me know if you think that's a good idea, or if you preferred I use a different version for this breaking change.

Regarding this, we've pretty much cemented the idea that version 2 of DiscordSRV is going to be [a full rewrite of the plugin](https://github.com/DiscordSRV/Ascension), so I don't think it's appropriate to use that as the version - even though just a minor bump feels wrong

src/main/java/github/scarsz/discordsrv/modules/voice/VoiceModule.java
```diff
@@ -177,14 +177,14 @@ private void tick() {
                     continue;
                 }
 
-                if (member.getVoiceState() == null || member.getVoiceState().getChannel() == null) {
+                if (member.getVoiceState() == null || member.getVoiceState().getChannel() == null || !(member.getVoiceState().getChannel() instanceof VoiceChannel)) {
```
Vankka, line 180: I'm wondering if it would be possible to support Stage channels with voice (for the voice lobby) - there is probably some common interface that has the appropriate methods

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -1416,22 +1417,33 @@ public void onDisable() {
...
+                                LangUtil.Message.CONSOLE_CHANNEL_TOPIC_AT_SERVER_SHUTDOWN.toString()
+                                        .replaceAll("%time%|%date%", time)
+                                        .replace("%serverversion%", serverVersion)
+                                        .replace("%totalplayers%", totalPlayers)
+                                        .replace("%timestamp%", shutdownTimestamp)
+                        );
+                    } else if (getConsoleChannel() != null) {
+                        warning("ChannelTopicUpdaterChannelTopicsAtShutdownEnabled is set, but the console channel " +
+                                "is a " + getConsoleChannel().getType() + " channel, and cannot be updated.");
+                    }
```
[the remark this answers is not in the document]
nightpool: wouldn't that just cause extra noise? Why log a warning for an option the user has explicitly disabled? 

My usecase is as follows: my main text channel goes to a thread. Threads cannot have their topics updated, so I disable the config var option for channel topic updating at startup and shutdown. Why should I see a warning in that case? What action would I take as an admin to get rid of that warning?

src/main/java/github/scarsz/discordsrv/modules/voice/VoiceModule.java
nightpool: Stage channels cannot be contained in a category, due to discord limitations, so the entire logic here would need a rework, correct? feels like this would be better as a follow-up PR/feature since it would change how the library works.

thewilloftheshadow: Stage channels can be in a category......

nightpool: hmm, maybe I was mistaken! I thought I remembered seeing that restriction in JDAs docs, but I must have misinterpreted it

src/main/java/github/scarsz/discordsrv/DiscordSRV.java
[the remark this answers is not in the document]
Vankka: I accidentally dropped the 'not' from that sentence, and I didn't notice that this was a part that already had a check for the topic updating being disabled as it just got left out of the diff. Ignore me

src/main/java/github/scarsz/discordsrv/modules/voice/VoiceModule.java
Vankka: The voice module shouldn't require the lobby to be in the category either way (although the documentation regarding that may be misleading)

granny:
All things considered I went ahead and rebased this PR onto the latest changes in the develop branch.

src/main/java/github/scarsz/discordsrv/modules/voice/VoiceModule.java
granny: Out of scope of this PR so I'm resolving this conversation.

Scarsz:
Hi @nightpool, thanks for going through the trouble of all this.

I've went ahead and merged your changes with upstream but from the time JDA 5 went from alpha to beta looks like it broke even more stuff.

If you have a chance, please give the branch another look, thank you

Vankka (approved):

Scarsz (changes requested):
Need to get compiling with latest JDA changes

PR #1402, opened 2022-05-10
Removed hidden RespectChatPlugins config option

BoredManCodes:


Vankka:
It's hidden on purpose

BoredManCodes:
But it isn't hidden on the Chinese translation. My PR removed it from that translation to bring it inline with the other languages

PR #1410, opened 2022-06-05
FIX: DiscordCommandFormat taking you to Discord's homepage sue to trailing period

BoredManCodes:


Vankka:
image
I cannot reproduce that and think that adding the space there would just make it ugly as hell

PR #1421, opened 2022-07-03
Slash Commands API

BlueTree242:
Hello, ive made an api for slash commands and permanently eliminate the issue where plugins override other's slash commands, so here is how it works:

Addons and plugins that hook into DiscordSRV adds their commands using `DiscordSRV.api.getSlashCommandsManager().addCommandsForPlugin(plugin, commands...)` and then when discordsrv finishes loading it will attempt register the total from all plugins, in case the plugin is disabled it wouldn't disable it, when the action is successful discordsrv sets the boolean `updated` so it does not re-register commands on next reload command if no plugin updated their commands, any plugin can reload slash commands whenever they wanted. The API also auto prints an error when the scopes are missing/rate limited


src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.stream.Collectors;
+
+@RequiredArgsConstructor
+public class SlashCommandManager {
+    @Getter
+    private boolean updated = false;
+    private final Map<Plugin, PluginCommands> commandsMap = new HashMap<>();
+
+    public void reloadSlashCommands() {
+        CommandListUpdateAction action = DiscordSRV.getPlugin().getMainGuild().updateCommands();
```
Vankka, line 47: I think it'd be a good idea to avoid using bulk update and instead adding, updating & deleting commands as necessary since the rate limit is per command that is added/updated/deleted so this would quickly drain the limit.

`getMainGuild` might return null. 

Also worth considering what use cases people have in mind, some times people add DiscordSRV to multiple servers, and might want to use slash commands in other guilds too

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,133 @@
...
+        DiscordSRV.getPlugin().getLogger().warning(plugins.size() == 1 && plugins.toArray(new Plugin[0])[0].equals(DiscordSRV.getPlugin()) ?
+                "DiscordSRV's Slash commands may not be registered" : //if discordsrv add commands in the future (like linkaccount command as far a
+                "Slash Commands for the following plugins may not be registered: " + plugins.stream().filter(Plugin::isEnabled).map(Plugin::getName).collect(Collectors.joining(", ")));
+        DiscordSRV.getPlugin().getLogger().warning("==============================================================");
+    }
+
+
+
+
+
```
Vankka, line 100:
```suggestion

```

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,133 @@
...
+        if (!added) return; //no need to register if no commands will be registered
+        action.queue(s -> updated = true,er -> printSlashRegistrationError(RegistrationResult.getResult(er), commandsMap.keySet()));
+    }
+
+    /**
+     * Sets the slash commands for a specific plugin, overrides old ones. this method does not register them, see {@link SlashCommandManager#reloadSlashCommands()}
+     * @param plugin Plugin that is registering the commands, whenever discordsrv is registering the commands it will not register the commands for that plugin if it is disabled
+     * @param commandData Commands to add
+     */
+    public void setCommandsForPlugin(@NotNull Plugin plugin, @NotNull CommandData... commandData) {
```
Vankka, line 64: This doesn't need a `ForPlugin` suffix. Would be nice to have add and get methods for this too

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private void printSlashRegistrationError(RegistrationResult result, Set<Plugin> plugins) {
+        DiscordSRV.getPlugin().getLogger().warning("==============================================================");
+        DiscordSRV.getPlugin().getLogger().warning("DiscordSRV could not register slash commands to your discord server!");
+        switch (result) {
+            case SUCCESS:
+                break;
+            case RATE_LIMIT:
+                DiscordSRV.getPlugin().getLogger().warning("Rate Limited! Are you restarting your server or reloading this plugin alot?");
+                break;
```
Vankka, line 81: I don't think such a big warning is required for being rate limited, 1 line should be enough

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
BlueTree242: But.. Rate limit is per request.. Not commands updated/edited/added, 100/guild/day it's per request

Vankka: image

[source](https://discord.com/developers/docs/interactions/application-commands#registering-a-command)
Not entirely sure about the scope of 'create' here, but non the less this is a daily limit

Xirado: This only includes unique updates because `updateCommands()` is idempotent for duplicate commands. Using `upsertCommand()` is a very bad idea, and should be avoided at all costs.

If you have 20 commands, you can use `updateCommands()` and set them all with only 1 request, instead of 20.

Vankka: Alright, so the way registering works has changed since I last used it, so that should actually be fine. The other comments still apply

Xirado: This hasn't changed. The only difference is with Application Permissions V2 where global commands now show up instantly instead of ~30 minutes.

BlueTree242: how do you want the big warning to look like when an error happens when registering the commands in a few servers? like server 1 missing scope, server 2 rate limited, server 3 unknown error... (in case you want registration in every guild)

LOOHP: Is the rate limit for 10 commands in a single `updateCommands()` call the same as 10 individual `upsertCommand()` calls?

Vankka: `<guild name>: <error message>` 1 line per server or something like that

BlueTree242:
everything's resolved

BlueTree242:
note that i did not test that api at all

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public void reloadSlashCommands() {
+        Set<CommandData> commands = new HashSet<>();
+        for (PluginCommands value : commandsMap.values()) {
+            if (value.plugin.isEnabled()) {
+                commands.addAll(value.data);
+            }
+        }
+        if (!commands.isEmpty()) return; //no need to register if no commands will be registered
```
freya022, line 60: If some commands were already registered before, but all of them were removed, the old commands would stay, leaving the bot's commands in an invalid state.

You should *still* update if there are no commands, which will remove the existing commands

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+        Map<Guild, RegistrationResult> errors = new HashMap<>();
+        for (Guild guild : DiscordSRV.getPlugin().getJda().getGuilds()) {
+            guild.updateCommands().addCommands(commands).queue(null, f -> errors.put(guild, RegistrationResult.getResult(f)));
```
freya022, line 63: `queue` is async, the `errors` map will never get populated fast enough for `printSlashRegistrationError` to do anything

You could replace the `queue` with a `submit`, store the `CompletableFuture` into a `List`, so then when all tasks are launched you can wait for all of them to finish

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Sets the slash commands for a specific plugin, overrides old ones. this method does not register them, see {@link SlashCommandManager#reloadSlashCommands()}
+     * @param plugin Plugin that is registering the commands, whenever discordsrv is registering the commands it will not register the commands for that plugin if it is disabled
+     * @param commandData Commands to add
+     */
+    public void setCommands(@NotNull Plugin plugin, @NotNull CommandData... commandData) {
+        PluginCommands pluginCommands = commandsMap.get(plugin);
+        if (pluginCommands == null) {
+            pluginCommands = new PluginCommands(plugin, new HashSet<>(Arrays.asList(commandData)));
+            commandsMap.put(plugin, pluginCommands);
```
freya022, line 77: Imo there's no need to reuse the PluginCommands object.

Make the `data` field final, and just replace the object in the map with the new set of commands. (i.e `commandsMap.put(plugin, new PluginCommands(plugin, commandData)`)

I think you should move the `HashSet` creation to the `PluginCommands` constructor

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+        } else pluginCommands.data = new HashSet<>(Arrays.asList(commandData));
+    }
+
+    private void printSlashRegistrationError(Map<Guild, RegistrationResult> errors, Set<Plugin> plugins) {
+        if (errors.size() == (int) errors.values().stream().filter(r -> r == RegistrationResult.RATE_LIMIT).count()) {
```
freya022, line 82: I have no idea how this means anything honestly

You should check if *at least one error is a ratelimit*

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public void reloadSlashCommands() {
+        Set<CommandData> commands = new HashSet<>();
+        for (PluginCommands value : commandsMap.values()) {
+            if (value.plugin.isEnabled()) {
+                commands.addAll(value.data);
+            }
+        }
+        if (!commands.isEmpty()) return; //no need to register if no commands will be registered
+        Map<Guild, RegistrationResult> errors = new HashMap<>();
```
freya022, line 61: The `Map` data structure is pointless here, you aren't going to get the values by key as you're simply iterating over all the values anyway.

You could just have a `List<RegistrationResult>` where the `RegistrationResult` also holds the Guild

**Don't forget to use a synchronized list**

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
```diff
@@ -0,0 +1,142 @@
...
+                    DiscordSRV.getPlugin().getLogger().warning("Missing scopes in " + guild.getName() + " (" + guild.getId() + ")");
+                case UNKNOWN_ERROR:
+                    DiscordSRV.getPlugin().getLogger().warning("Unknown error in " + guild.getName() + " (" + guild.getId() + ")");
+                case RATE_LIMIT:
+                    DiscordSRV.getPlugin().getLogger().warning("Rate limited in " + guild.getName() + " (" + guild.getId() + ")");
+            }
+        }
+        if (errors.values().stream().anyMatch(r -> r == RegistrationResult.MISSING_SCOPE)) DiscordSRV.getPlugin().getLogger().warning("Use " + invite + " to re-invite the bot to guilds with missing scope!");
+        DiscordSRV.getPlugin().getLogger().warning(" ");
+        DiscordSRV.getPlugin().getLogger().warning(plugins.size() == 1 && plugins.toArray(new Plugin[0])[0].equals(DiscordSRV.getPlugin()) ?
```
freya022, line 103:
```suggestion
        DiscordSRV.getPlugin().getLogger().warning(plugins.size() == 1 && plugins.get(0).equals(DiscordSRV.getPlugin()) ?
```

And stop spamming `DiscordSRV.getPlugin().getLogger()`

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
BlueTree242: the problem with that is that there might be no addons on the server to register any slash commands, and if there is no scope provided the errors would show up and user wouldn't care and it would annoy them so much

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
BlueTree242: lmao how did i miss that 😅

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
freya022: Yes but the check should check if there's at least one rate limit

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
freya022: Then you just have to check if there are any commands being registered, when printing the missing scope error

freya022: Actually, why not just remove this one check, since you're going to print rate limited guilds below anyway ?

src/main/java/github/scarsz/discordsrv/api/SlashCommandManager.java
freya022: Actually, why not just remove this one check, since you're going to print rate limited guilds below anyway ?

BlueTree242: reduce the error if it's all just rate limit on all, it was vankka's request

BlueTree242:
everything should be good now

Vankka (approved):

PR #1424, opened 2022-07-08
Add Staff++ vanish hook

garagepoort:


Vankka:
I've been trying to move away from doing per plugin vanish hooks, and towards using the `vanished` metadata key. The metadata key is already being used by [VanishNoPacket](https://github.com/mbax/VanishNoPacket/blob/e0133f30708a340df76e4f24517d4e532c5a294a/src/main/java/org/kitteh/vanish/VanishPlugin.java#L290), [SuperVanish](https://github.com/LeonMangler/SuperVanish/blob/e6e3782b04ee808d5a231516be31c09cebea01f8/src/main/java/de/myzelyam/supervanish/visibility/VisibilityChanger.java#L74) and [Essentials](https://github.com/EssentialsX/Essentials/blob/6816eb4e180cb4114ee982ef7400acba5b18d181/Essentials/src/main/java/com/earth2me/essentials/User.java#L968) (some of the vanish hooks are from days before the metadata key)

https://github.com/DiscordSRV/DiscordSRV/blob/339e5e81da67cfe6210679af775eb66a9b5033be/src/main/java/github/scarsz/discordsrv/DiscordSRV.java#L1206
I would suggest that Staff++ should start using the metadata key

garagepoort:
> I've been trying to move away from doing per plugin vanish hooks, and towards using the `vanished` metadata key. The metadata key is already being used by [VanishNoPacket](https://github.com/mbax/VanishNoPacket/blob/e0133f30708a340df76e4f24517d4e532c5a294a/src/main/java/org/kitteh/vanish/VanishPlugin.java#L290), [SuperVanish](https://github.com/LeonMangler/SuperVanish/blob/e6e3782b04ee808d5a231516be31c09cebea01f8/src/main/java/de/myzelyam/supervanish/visibility/VisibilityChanger.java#L74) and [Essentials](https://github.com/EssentialsX/Essentials/blob/6816eb4e180cb4114ee982ef7400acba5b18d181/Essentials/src/main/java/com/earth2me/essentials/User.java#L968) (some of the vanish hooks are from days before the metadata key)
> 
> https://github.com/DiscordSRV/DiscordSRV/blob/339e5e81da67cfe6210679af775eb66a9b5033be/src/main/java/github/scarsz/discordsrv/DiscordSRV.java#L1206
> 
> 
> I would suggest that Staff++ should start using the metadata key

Ah I didn't see that. Will do

Vankka:
Nice, sorry about it being a bit unclear, I went ahead and made the VanishHook related stuff deprecated https://github.com/DiscordSRV/DiscordSRV/commit/5a051d31886683f4345893adf70ff826e53cd6b8 to make it more clear

PR #1431, opened 2022-07-31
Interface-based Slash Commands API

Scarsz:
TODO:

- [x] add manual command handler registration
- [x] annotation for slash command providers to denote command handlers
- [x] allow commands to be guild-specific (@LOOHP)
- [x] proper error logging of unknown/unexpected errors

Used by implementing `SlashCommandProvider`. Plugins implementing said interface are automatically registered. All a plugin needs to do is implement the `SlashCommandProvider#getSlashCommands` method to provide command data and provide @SlashCommand-marked methods to automatically be called.

```java
public class YourPlugin extends JavaPlugin implements SlashCommandProvider {

    @Override
    public Set<PluginSlashCommand> getSlashCommands() {
        return new HashSet<>(Arrays.asList(
                // ping pong
                new PluginSlashCommand(this, new CommandData("ping", "A classic match of ping pong")),

                // economy commands
                new PluginSlashCommand(this, new CommandData("eco", "Economy commands").addSubcommands(
                        new SubcommandData("stimulus", "Grant all online players a cash bonus")
                                .addOption(OptionType.INTEGER, "amount", "Amount of money to grant online players", true),
                        new SubcommandData("capsize", "Capsize your economy in one simple command")
                )),

                // most important command
                new PluginSlashCommand(this, new CommandData("life", "What is the meaning of life?"))
        ));
    }

    @SlashCommand(path = "ping")
    public void pingCommand(SlashCommandEvent event) {
        event.reply("Pong!").queue();
    }

    @SlashCommand(path = "eco/stimulus")
    public void ecoStimulus(SlashCommandEvent event) {
        double amount = event.getOption("amount").getAsDouble();
        for (Player player : Bukkit.getOnlinePlayers()) economy.addMoney(player, amount);
        event.reply("✅ Granted " + Bukkit.getOnlinePlayers().size() + " players $" + amount).queue();
    }
    @SlashCommand(path = "eco/capsize")
    public void ecoCapsizeCommand(SlashCommandEvent event) {
        for (Player player : Bukkit.getOnlinePlayers()) economy.addMoney(player, 1_000_000_000);
        event.reply("✅ Economy destroyed!").queue();
    }

    @SlashCommand(path = "life")
    public void meaningOfLifeCommandImplicitDefer(SlashCommandEvent event) {
        event.deferReply() // must defer reply if command processing will take longer than 3 seconds
                .delay(5, TimeUnit.SECONDS) // heavy thinking
                .map(hook -> hook.sendMessage("42"))
                .queue();
    }
    // or tell DiscordSRV to do the deferring for you if you know your command will take a while...
    @SlashCommand(path = "life", deferReply = true)
    public void meaningOfLifeCommandExplicitDefer(SlashCommandEvent event) {
        event.getHook().sendMessage("42").queueAfter(5, TimeUnit.SECONDS);
    }

}
```

If a plugin does not wish to hard-depend on DiscordSRV, they can manually add a `SlashCommandProvider` instance via `DiscordSRV.api.addSlashCommandProvider(this)`.


LOOHP:
The ability to control which guilds to register a command would be nice.

Scarsz:
> The ability to control which guilds to register a command would be nice.

@LOOHP Implemented

LOOHP:
> > The ability to control which guilds to register a command would be nice.
> 
> @LOOHP Implemented

Thanks, looks great!

src/main/java/github/scarsz/discordsrv/api/ApiManager.java
```diff
@@ -124,55 +137,195 @@ public <E extends Event> E callEvent(E event) {
...
+                if (!slashCommand.deferReply()) {
+                    invokeMethod(method, provider, event);
+                } else {
+                    event.deferReply(slashCommand.deferEphemeral())
+                            .queue(hook -> invokeMethod(method, provider, event));
+                }
+
+                if (!event.isAcknowledged()) {
+                    DiscordSRV.error(String.format(
+                            "Slash command for \"/%s\" was not acknowledged by %s's handler! The command will show as failed on Discord until this is fixed!",
```
Vankka, line 270:
```suggestion
                            "Slash command \"/%s\" was not acknowledged by %s's handler! The command will show as failed on Discord until this is fixed!",
```

src/main/java/github/scarsz/discordsrv/api/commands/PluginCommandData.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Set;
+
+/**
+ * {@link CommandData} wrapper that includes the originating {@link Plugin}
+ */
+@Value
+public class PluginCommandData extends CommandData {
+
+    Plugin plugin;
+    CommandData commandData;
```
Vankka, line 20: The `CommandData` field isn't being used besides in `equals`, this is misleading as users might construct a CommandData object then pass it to PluginCommandData thinking it's details (eg. anything besides name and description) would be used, when that isn't the case.

Alternatively this class could be changed to not extend CommandData and use the field's CommandData instead.

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommand.java
```diff
@@ -0,0 +1,39 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface SlashCommand {
+
+    /**
+     * The command path that shall be handled by this method
+     * @return the command path
+     * @see SlashCommandEvent#getCommandPath()
+     */
+    String path();
```
Vankka, line 24: Example here would be nice, explaining what parts of the command (command + subcommand group + subcommand) are included and how they need to be formatted (separated by slashes)

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommand.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Tells DiscordSRV to automatically acknowledge the command & defer replying for you.
+     * <p>
+     * <strong>You cannot acknowledge the command yourself if you enable this.</strong>
+     * Send your replies through the {@link InteractionHook} via <code>{@link SlashCommandEvent}.getHook().sendMessage(...)</code>
+     * @return whether DiscordSRV should automatically defer the interaction's reply when routing the event to your handler
+     */
+    boolean deferReply() default false;
```
Vankka, line 33: Should note why deferring might be useful (3? second responding time vs what deferred replies have)

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandProvider.java
```diff
@@ -0,0 +1,24 @@
...
+import net.dv8tion.jda.api.interactions.commands.CommandInteraction;
+import org.bukkit.plugin.Plugin;
+
+import java.util.Set;
+
+/**
+ * Provides {@link PluginCommandData} datum to DiscordSRV's {@link ApiManager}.
+ * {@link Plugin}s implementing this interface are automatically registered.
+ * To manually register a provider, use {@link ApiManager#addSlashCommandProvider(SlashCommandProvider)}.
+ * To handle commands, create a method in a {@link SlashCommandProvider} with the {@link SlashCommand} annotation.
```
Vankka, line 14: Had to lookup what datum means, and doesn't seem too fitting there.

Should note the required parameter for the method when using the annotation

Vankka (changes requested):

Vankka (approved):

Scarsz:
Had to switch from extending `CommandData` to accepting/storing/passing a built `CommandData` instance. The intermediary methods on `CommandData` like `#addSubcommands` return `CommandData` instead of `<? extends CommandData>` thus prevent chaining because of type mismatches

Vankka (approved):

PR #1433, opened 2022-08-01
Link accounts via guild text channel

Kamesuta:
# Link accounts via guild text channel

Because the process of sending a PM to the Discord Bot to link accounts is complicated, I have added a way to link by sending a code to the text channel.

## How to enable the feature
1. See the `config.yml`
2. Add the channel id or name to `MinecraftDiscordAccountChannels`
3. Well done! You can type link code to the text channel

## Extra change
- Changed reply message to link code to use Discord's reply feature

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
@@ -36,7 +36,6 @@ import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import org.bukkit.Bukkit;
 import org.bukkit.OfflinePlayer;
 
-import java.util.Objects;
 import java.util.UUID;
 
 public class DiscordAccountLinkListener extends ListenerAdapter {
@@ -62,7 +61,7 @@ public class DiscordAccountLinkListener extends ListenerAdapter {
 
         // if message is not in the link channel, don't process it
         TextChannel linkChannel = DiscordSRV.getPlugin().getDestinationTextChannelForGameChannelName("link");
-        if (!Objects.equals(event.getChannel(), linkChannel)) return;
+        if (!event.getChannel().equals(linkChannel)) return;
 
         String reply = DiscordSRV.getPlugin().getAccountLinkManager().process(event.getMessage().getContentRaw(), event.getAuthor().getId());
         if (reply != null) event.getMessage().reply(reply).queue();
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
@@ -24,11 +24,11 @@ package github.scarsz.discordsrv.listeners;
 
 import github.scarsz.discordsrv.Debug;
 import github.scarsz.discordsrv.DiscordSRV;
-import github.scarsz.discordsrv.api.events.DiscordGuildMessageReceivedEvent;
 import github.scarsz.discordsrv.api.events.DiscordPrivateMessageReceivedEvent;
 import github.scarsz.discordsrv.util.DiscordUtil;
 import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.entities.Role;
+import net.dv8tion.jda.api.entities.TextChannel;
 import net.dv8tion.jda.api.events.guild.member.GuildMemberJoinEvent;
 import net.dv8tion.jda.api.events.message.guild.GuildMessageReceivedEvent;
 import net.dv8tion.jda.api.events.message.priv.PrivateMessageReceivedEvent;
@@ -36,7 +36,7 @@ import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import org.bukkit.Bukkit;
 import org.bukkit.OfflinePlayer;
 
-import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 
 public class DiscordAccountLinkListener extends ListenerAdapter {
@@ -49,7 +49,7 @@ public class DiscordAccountLinkListener extends ListenerAdapter {
         DiscordSRV.api.callEvent(new DiscordPrivateMessageReceivedEvent(event));
 
         // don't link accounts if config option is disabled
-        if (!DiscordSRV.config().getBoolean("MinecraftDiscordAccountEnablePM")) return;
+        if (!DiscordSRV.config().getBoolean("MinecraftDiscordAccountLinkedUsePM")) return;
 
         String reply = DiscordSRV.getPlugin().getAccountLinkManager().process(event.getMessage().getContentRaw(), event.getAuthor().getId());
         if (reply != null) event.getMessage().reply(reply).queue();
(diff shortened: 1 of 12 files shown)


src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
```diff
@@ -45,8 +48,28 @@ public void onPrivateMessageReceived(PrivateMessageReceivedEvent event) {
...
+        String reply = DiscordSRV.getPlugin().getAccountLinkManager().process(event.getMessage().getContentRaw(), event.getAuthor().getId());
+        if (reply != null) event.getMessage().reply(reply).queue();
+    }
+
+    @Override
+    public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
+        // don't process messages sent by the bot
+        if (event.getAuthor().getId().equals(event.getJDA().getSelfUser().getId())) return;
+
+        DiscordSRV.api.callEvent(new DiscordGuildMessageReceivedEvent(event));
```
Scarsz, line 63: This event is already published by the chat listener, it shouldn't be published twice

src/main/resources/config/en.yml
```diff
@@ -286,11 +286,15 @@ DiscordCannedResponses: {"!ip": "yourserveripchange.me", "!site": "http://yoursi
...
 # MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: allows sending a new code to the bot to unlink and relink with the new code
+# MinecraftDiscordAccountEnablePM: allows linking accounts using PM
+# MinecraftDiscordAccountChannels: the name or id of a discord channel to link accounts
 #
 MinecraftDiscordAccountLinkedConsoleCommands: ["", "", ""]
 MinecraftDiscordAccountUnlinkedConsoleCommands: ["", "", ""]
 MinecraftDiscordAccountLinkedRoleNameToAddUserTo: "Linked"
 MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: false
+MinecraftDiscordAccountEnablePM: true
+MinecraftDiscordAccountChannels: []
```
Scarsz, line 297: We already have a `Channels` option, I feel it would be fine to just allow linking in the already defined channels

Scarsz (changes requested):

src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
Kamesuta: Okay, I removed it

src/main/resources/config/en.yml
Kamesuta: I would like to have a dedicated channel for linking on my server and guide people to link on the channel for linking.

So I will try to implement an additional option to link when writing to Channels as well.

Scarsz: I'd advise adding a "link account in any registered chat channel" option and looking up a `link` named channel in the `Channels` config for a dedicated link channel otherwise

Kamesuta: I understand. I'll try.

Kamesuta: Okay, thanks

Kamesuta: If so, how would you specify if you wanted to limit it to PMs only?

Scarsz: `MinecraftDiscordAccountEnablePM: true` is fine to me as-is

Kamesuta: Ok,
`MinecraftDiscordAccountEnablePM: true` → Only PMs are allowed to link.
`MinecraftDiscordAccountEnablePM: false` → Only text channels are allowed to link.

Scarsz: Channel linking should be available at the same time PM linking is, the PM linking option should just control whether people can/can't link with a PM

Kamesuta: You are right.
I have found that I cannot mix a channel for chat and a link channel.
This is because the feature of replying with an invalid link code warning interferes with the ability to chat.
Therefore, the link channel and the other channels need to be separated.

So I'll implement the following
`MinecraftDiscordAccountEnablePM: true/false` → enable/disable PMs to link.
`Channels`.`link` → empty for disable, fill channel id for enable text channels to link.

Kamesuta: I fixed it.

src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
```diff
@@ -45,8 +48,24 @@ public void onPrivateMessageReceived(PrivateMessageReceivedEvent event) {
...
+    }
+
+    @Override
+    public void onGuildMessageReceived(GuildMessageReceivedEvent event) {
+        // don't process messages sent by the bot
+        if (event.getAuthor().getId().equals(event.getJDA().getSelfUser().getId())) return;
+
+        // if message is not in the link channel, don't process it
+        TextChannel linkChannel = DiscordSRV.getPlugin().getDestinationTextChannelForGameChannelName("link");
+        if (!Objects.equals(event.getChannel(), linkChannel)) return;
```
Scarsz, line 65: This should just be `!event.getChannel().equals(linkChannel)`

Scarsz (changes requested):

src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
Kamesuta: I fixed it.

BoredManCodes:
There's an addon for this btw https://www.spigotmc.org/resources/discordsrv-linkchannel.101604/

src/main/resources/config/fr.yml
```diff
@@ -283,11 +284,14 @@ DiscordCannedResponses: {"!ip": "yourserveripchange.me", "!site": "http://yoursi
...
 # MinecraftDiscordAccountLinkedSetDiscordNicknameAsInGameName: le nom discord du joueur doit-il être modifié par son nom Minecraft ?
 # MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: permet d'envoyer un nouveau code au bot pour dissocier et relier le nouveau code
+# MinecraftDiscordAccountLinkedUsePM: Lier des comptes à l'aide de PMs
 #
 MinecraftDiscordAccountLinkedConsoleCommands: ["", "", ""]
 MinecraftDiscordAccountUnlinkedConsoleCommands: ["", "", ""]
 MinecraftDiscordAccountLinkedRoleNameToAddUserTo: "Linked"
 MinecraftDiscordAccountLinkedAllowRelinkBySendingANewCode: false
+MinecraftDiscordAccountLinkedUsePM: true
+
```
Vankka, line 294: Extra line
```suggestion
```

src/main/resources/config/en.yml
```diff
@@ -23,6 +23,7 @@ BotToken: "BOTTOKEN"
 # for dynmap messages: dynmap
 # for watchdog messages: watchdog
 # for /discord broadcast: broadcasts (unless specified in the command)
+# for account link: link
```
Vankka, line 26:
```suggestion
# for account linking: link
```

src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
```diff
@@ -45,8 +47,24 @@ public void onPrivateMessageReceived(PrivateMessageReceivedEvent event) {
...
+        // don't process messages sent by the bot
+        if (event.getAuthor().getId().equals(event.getJDA().getSelfUser().getId())) return;
+
+        // if message is not in the link channel, don't process it
+        TextChannel linkChannel = DiscordSRV.getPlugin().getDestinationTextChannelForGameChannelName("link");
+        if (!event.getChannel().equals(linkChannel)) return;
+
         String reply = DiscordSRV.getPlugin().getAccountLinkManager().process(event.getMessage().getContentRaw(), event.getAuthor().getId());
-        if (reply != null) event.getChannel().sendMessage(reply).queue();
+        if (reply != null) event.getMessage().reply(reply).queue();
```
Vankka, line 67: Would be nice to have a option to automatically delete the request and response (after 5 seconds or a different configured value) to keep the channel clean

Vankka (changes requested):

src/main/resources/config/fr.yml
Kamesuta: Sorry, I made a mistake.
I fixed it.

src/main/resources/config/en.yml
Kamesuta: I fixed it.

src/main/java/github/scarsz/discordsrv/listeners/DiscordAccountLinkListener.java
Kamesuta: Ok, I implemented a feature to delete after a configured number of seconds

Vankka (approved):

PR #1438, opened 2022-08-12
Schedule slash commands to be updated later when plugins are ready

LOOHP:
This change makes it so that plugins won't have to register `onLoad()` if they want to catch the initial update slash command call.


Scarsz (commented):
Is the delay really necessary? Should probably be 0

LOOHP:
Not really, I just used the delay used by server start messages.

LOOHP:
Not really, I just used the delay used by server start messages. I'll just use runTask

PR #1439, opened 2022-08-14
Patch slash command api bugs

LOOHP:
Fixed and improved a few things for the slash command api
- `null` should be of type String[] in the constructor of `PluginSlashCommand`
- Prevent double `SlashCommandProvider` registration, if an instance is already a plugin main class instance
- `SlashCommandEvent` is not called for plugin main class instances of `SlashCommandProvider`s
- Only warn a slash command is not acknowledged by a plugin if none of its listeners responded to it, instead of once for each of them that doesn't
- Remove guild filter methods for `PluginSlashCommand` for chaining

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
+++ b/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
@@ -247,6 +247,12 @@ public class ApiManager extends ListenerAdapter {
      * @param provider the command data provider
      */
     public void addSlashCommandProvider(@NonNull SlashCommandProvider provider) {
+        // Prevent double registration, if an instance is already a plugin instance
+        for (Plugin plugin : Bukkit.getPluginManager().getPlugins()) {
+            if (provider == plugin) {
+                return;
+            }
+        }
         this.slashCommandProviders.add(provider);
     }
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/github/scarsz/discordsrv/api/ApiManager.java
```diff
@@ -247,6 +247,12 @@ public void updateSlashCommands() {
      * @param provider the command data provider
      */
     public void addSlashCommandProvider(@NonNull SlashCommandProvider provider) {
+        // Prevent double registration, if an instance is already a plugin instance
+        for (Plugin plugin : Bukkit.getPluginManager().getPlugins()) {
```
Scarsz, line 251: Check if the provider is even a plugin before looping through all the plugins

src/main/java/github/scarsz/discordsrv/api/ApiManager.java
```diff
@@ -247,6 +247,12 @@ public void updateSlashCommands() {
...
+            if (provider == plugin) {
+                return;
+            }
+        }
         this.slashCommandProviders.add(provider);
```
Vankka, line 256: `slashCommandProviders` is a set that doesn't accept duplicates, this check is unnecessary

src/main/java/github/scarsz/discordsrv/api/ApiManager.java
```diff
@@ -247,6 +247,12 @@ public void updateSlashCommands() {
      * @param provider the command data provider
      */
     public void addSlashCommandProvider(@NonNull SlashCommandProvider provider) {
+        // Prevent double registration, if an instance is already a plugin instance
+        for (Plugin plugin : Bukkit.getPluginManager().getPlugins()) {
+            if (provider == plugin) {
+                return;
```
Scarsz, line 253: Short circuits should just be one line

Vankka (changes requested):
Single comment below, otherwise looks good. Thanks

Scarsz (changes requested):

PR #1441, opened 2022-08-15
Fire an event before responding to player list commands

LOOHP:
Added new `DiscordChatChannelListCommandMessageEvent`, this event is fired before responding to player list commands by discordsrv, allowing plugins to modify the response.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/api/events/DiscordChatChannelListCommandMessageEvent.java
+++ b/src/main/java/github/scarsz/discordsrv/api/events/DiscordChatChannelListCommandMessageEvent.java
@@ -50,12 +50,13 @@ public class DiscordChatChannelListCommandMessageEvent extends Event {
     @Getter @Setter private String playerListMessage;
     @Getter @Setter private int expiration;
 
-    public DiscordChatChannelListCommandMessageEvent(TextChannel channel, Guild guild, String message, GuildMessageReceivedEvent triggeringJDAEvent, String playerListMessage, Result result) {
+    public DiscordChatChannelListCommandMessageEvent(TextChannel channel, Guild guild, String message, GuildMessageReceivedEvent triggeringJDAEvent, String playerListMessage, int expiration, Result result) {
         this.channel = channel;
         this.guild = guild;
         this.message = message;
         this.triggeringJDAEvent = triggeringJDAEvent;
         this.playerListMessage = playerListMessage;
+        this.expiration = expiration;
         this.result = result;
     }
 
@@ -72,7 +73,7 @@ public class DiscordChatChannelListCommandMessageEvent extends Event {
 
         /**
          * <p>Cancel the event, treat message as regular discord message for further processing.</p>
-         * <p>(For example, be boardcasted in game)</p>
+         * <p>(For example, be broadcasted in game)</p>
          */
         TREAT_AS_REGULAR_MESSAGE;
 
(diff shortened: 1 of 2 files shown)


src/main/java/github/scarsz/discordsrv/api/events/DiscordChatChannelListCommandMessageEvent.java
```diff
@@ -0,0 +1,81 @@
...
+        SEND_RESPONSE,
+
+        /**
+         * <p>Cancel the event, do not perform any action afterwards.</p>
+         */
+        NO_ACTION,
+
+        /**
+         * <p>Cancel the event, treat message as regular discord message for further processing.</p>
+         * <p>(For example, be boardcasted in game)</p>
```
Scarsz, line 75: Typo

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -407,10 +407,11 @@ private boolean processPlayerListCommand(GuildMessageReceivedEvent event, String
         if (!DiscordSRV.config().getBoolean("DiscordChatChannelListCommandEnabled")) return false;
         if (!StringUtils.trimToEmpty(message).equalsIgnoreCase(DiscordSRV.config().getString("DiscordChatChannelListCommandMessage"))) return false;
 
+        int expiration = DiscordSRV.config().getInt("DiscordChatChannelListCommandExpiration") * 1000;
+        String playerListMessage = "";
```
Scarsz, line 411: Is the initial value really necessary?

Scarsz (changes requested):

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
LOOHP: Not really, it was just left there from before I made any changes.

Scarsz (approved):

Scarsz:
Realistically this needs to be turned into a slash command eventually

LOOHP:
> Realistically this needs to be turned into a slash command eventually

Agreed, I've just added this event because some users still prefer the legacy text-based command and wish that the `/playerlist` response from my addon would also apply to this.

If DiscordSRV decides to implement `/playerlist` then there will need to be a way for plugins to override them.

LOOHP:
If there are no plans to remove the old text-based `playerlist` command, then this can be merged, correct?

src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
```diff
@@ -444,17 +445,29 @@ private boolean processPlayerListCommand(GuildMessageReceivedEvent event, String
...
+
+                // expire message after specified time
+                if (listCommandMessageEvent.getExpiration() > 0 && DiscordSRV.config().getBoolean("DiscordChatChannelListCommandExpirationDeleteRequest")) {
+                    new Thread(() -> {
+                        try {
+                            Thread.sleep(listCommandMessageEvent.getExpiration());
+                        } catch (InterruptedException ignored) {
+                        }
+                        DiscordUtil.deleteMessage(event.getMessage());
+                    }).start();
```
Vankka, line 464: Should use queueAfter instead of sleeping a thread for such a little operation

Vankka (changes requested):

Vankka (approved):

PR #1442, opened 2022-08-16
Slash Command Priorities

LOOHP:
This PR introduces slash command priorities both during command registration and command handling. 
It opens the door to converting DiscordSRV's built-in text commands into slash commands while allowing other plugins to override them.

Tho there are a few questions
- Is this a good way to approach this problem?
- If so, should `SlashCommandPriority` in `SlashCommand` annotations and `PluginSlashCommand` fields use separated enums?

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
+++ b/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
@@ -217,7 +217,10 @@ public class ApiManager extends ListenerAdapter {
         int finalConflictingCommands = conflictingCommands;
         RestAction.allOf(guildCommandUpdateActions).queue(all -> {
             int successful = all.stream().filter(Objects::nonNull).mapToInt(List::size).sum();
-            DiscordSRV.info("Successfully registered " + successful + " slash commands (" + finalConflictingCommands + " conflicted) for " + conflictResolvedCommands.values().stream().map(PluginSlashCommand::getPlugin).distinct().count() + " plugins in " + all.stream().filter(Objects::nonNull).count() + "/" + DiscordSRV.getPlugin().getJda().getGuilds().size() + " guilds (" + finalCancelledGuilds + " cancelled)");
+            long pluginCount = conflictResolvedCommands.values().stream().map(PluginSlashCommand::getPlugin).distinct().count();
+            long registeredGuilds = all.stream().filter(Objects::nonNull).count();
+            int totalGuilds = DiscordSRV.getPlugin().getJda().getGuilds().size();
+            DiscordSRV.info("Successfully registered " + successful + " slash commands (" + finalConflictingCommands + " conflicted) for " + pluginCount + " plugins in " + registeredGuilds + "/" + totalGuilds + " guilds (" + finalCancelledGuilds + " cancelled)");
 
             if (errors.isEmpty()) return;
 
--- a/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
+++ b/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
@@ -184,7 +184,7 @@ public class ApiManager extends ListenerAdapter {
             PluginSlashCommand conflictingCommand = conflictResolvedCommands.putIfAbsent(name, pluginSlashCommand);
             if (conflictingCommand == null) continue;
             conflictingCommands++;
-            if (pluginSlashCommand.getPriority().ordinal() > conflictingCommand.getPriority().ordinal())
+            if (pluginSlashCommand.getPriority().ordinal() < conflictingCommand.getPriority().ordinal())
                 conflictResolvedCommands.put(name, pluginSlashCommand);
         }
 
--- a/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
+++ b/src/main/java/github/scarsz/discordsrv/api/ApiManager.java
@@ -28,6 +28,7 @@ import github.scarsz.discordsrv.DiscordSRV;
 import github.scarsz.discordsrv.api.commands.CommandRegistrationError;
 import github.scarsz.discordsrv.api.commands.PluginSlashCommand;
 import github.scarsz.discordsrv.api.commands.SlashCommand;
+import github.scarsz.discordsrv.api.commands.SlashCommandPriority;
 import github.scarsz.discordsrv.api.commands.SlashCommandProvider;
 import github.scarsz.discordsrv.api.events.Event;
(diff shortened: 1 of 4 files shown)


src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
```diff
@@ -0,0 +1,66 @@
...
+     * <p>When in {@link SlashCommand} - Slash command handler is not important and may give way to other handlers who wishes to overtake</p>
+     */
+    LATE(3),
+    /**
+     * <p>When in {@link PluginSlashCommand} - Slash command should be overridden by any other if other conflicting commands exist</p>
+     * <p>When in {@link SlashCommand} - Slash command handler should always be overtaken if any other handlers wishes to </p>
+     */
+    LAST(4);
+
+    @Getter private final int slot;
```
Scarsz, line 60: The Enum#ordinal value should just be used instead of a dedicated slot id

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
```diff
@@ -0,0 +1,66 @@
...
+package github.scarsz.discordsrv.api.commands;
+
+import lombok.Getter;
+
+/**
+ * <p>Used to indicate the registration priority of slash commands and the invocation order of slash command handlers</p>
+ * <p>Defaults to {@link #NORMAL} in {@link PluginSlashCommand} where it's not specifically set</p>
+ * <p>Defaults to {@link #NORMAL} in {@link SlashCommand} annotations where it's not specifically set</p>
+ */
+public enum SlashCommandPriority {
```
Scarsz, line 32: Is there a reason that the existing API's EventPriority enum couldn't be used? It's enum values are more self explanatory as well

Scarsz (dismissed):

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
LOOHP: In my opinion, using "Listener" priorities to indicate the registration order of slash commands seems a bit weird and misleading, since they aren't listeners. The existing java doc in the listener priority class would also be very lengthy if we have to add two more uses.

Scarsz: Forgot that Enum's name is ListenerPriority. In my opinion that enum should be deprecated in favor of a generic enum like EventPriority. The documentation wouldn't really be an issue.

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
Scarsz: Pretty sure that's the same way Bukkit's API does it internally

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
LOOHP: There is also another problem, for slash command handlers, an "important" handler should run first, as opposed to bukkit events where "important" handlers run last. Using the same enum for both cases would be very confusing.

src/main/java/github/scarsz/discordsrv/api/ApiManager.java
```diff
@@ -174,15 +177,26 @@ public void updateSlashCommands() {
...
+        int conflictingCommands = 0;
+        Map<String, PluginSlashCommand> conflictResolvedCommands = new HashMap<>();
+        for (PluginSlashCommand pluginSlashCommand : commands) {
+            String name = pluginSlashCommand.getCommandData().getName();
+            PluginSlashCommand conflictingCommand = conflictResolvedCommands.putIfAbsent(name, pluginSlashCommand);
+            if (conflictingCommand == null) continue;
+            conflictingCommands++;
+            if (pluginSlashCommand.getPriority().ordinal() < conflictingCommand.getPriority().ordinal())
+                conflictResolvedCommands.put(name, pluginSlashCommand);
+        }
```
Vankka, line 189: Non conflicting commands are never actually added to `conflictResolvedCommands`

src/main/java/github/scarsz/discordsrv/api/commands/PluginSlashCommand.java
```diff
@@ -107,5 +128,14 @@ public PluginSlashCommand removeGuildFilter(String guildId) {
         this.guilds.remove(guildId);
         return this;
     }
+    /**
+     * Set the priority of this slash command when handling registration conflicts
+     * @param priority the priority of this slash command
+     * @return the {@link PluginSlashCommand} instance for chaining
+     */
+    public PluginSlashCommand priority(SlashCommandPriority priority) {
```
Vankka, line 136:
```suggestion
    public PluginSlashCommand setPriority(SlashCommandPriority priority) {
```

Vankka (changes requested):

src/main/java/github/scarsz/discordsrv/api/ApiManager.java
LOOHP: That should be correct, `conflictResolvedCommands` is the set of commands from `commands` but with the conflicting ones removed according to their priority.

Vankka: Nevermind me I thought putIfAbsent was using a different collection

LOOHP:
_Hmm why does it still shows "2 reviews requesting changes"?_

Vankka:
It will only stop showing that if both mine and Scarsz' reviews are either dismissed or we approve the changes overwriting the previous request changes review

src/main/java/github/scarsz/discordsrv/api/commands/SlashCommandPriority.java
Vankka: The only way to make changes to ListenerPriority would be to break the API or make a annotation to replace `@Subscribe`. I'm in favor of moving toward a single more generic enum like "Order", though

LOOHP: `@Subscribe` is a pretty big part of the API though, is now the right time to break it? I feel like DiscordSRV v2 would be better timing for that.

Vankka (approved):

Scarsz:
lgtm

PR #1459, opened 2022-09-18
Add PlaceholderAPI support to account linking code messages

Androkai:
As requested by François#7192 in support ticket 11134, adds support for PlaceholderAPI usage in DiscordCommandFormat message.


Vankka (approved):

R2D2VaderBeef:
Just a clarification, but I believe that is actually the CodeGenerated message, as it includes a linking code and linking instructions (see attached screenshot) 
image

Androkai:
> Just a clarification, but I believe that is actually the CodeGenerated message they are requesting, as it includes a linking code and linking instructions (see attached screenshot) image

Could have sworn that was at least once included directly in the original message, but you're right, I'll add that to the message quickly.

Vankka (approved):

PR #1471, opened 2022-10-12
Add typing indicators

### LINKED ISSUE #1412: Typing indication
### Why?

When a player begins typing in the Minecraft chat, have the bot begin typing in the main chat channel.

### What and How?

With 1.19's addition of `previews-chat` in the server properties, if enabled and agreed to by players, everything the player types is sent to the server before the player actually finishes the message.

On receiving these events, a the server may enqueue a typing indication for discord.

### Are there alternatives?

For 1.16-1.18.2 I had been using a tiny plugin that listens for "typing" event on a custom channel, and also made a tiny fabric mod that sent that same "typing" event for every player keypress where the chat text entry is not empty and isn't a command. This alternative worked out alright but of course required players to actually install the fabric mod to participate.

### Checks

### Anything else

_No response_

ConsoleLogLuke:
This adds an optional feature that makes the bot start "typing" in Discord while players are typing in Minecraft, as described in #1412. It does this by taking advantage of the new [`AsyncPlayerChatPreviewEvent`](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/AsyncPlayerChatPreviewEvent.html) that was added in Spigot 1.19.1.

It adds a config option, `DiscordChatChannelTypingIndicators`, which is disabled by default and does nothing on versions prior to 1.19.1 (except sent a warning message to the console). The `previews-chat` option also needs to be enabled in server.properties or a warning message will be sent to the console.

If a chat hook is enabled, typing indicators will be correctly sent for the channel the player is currently talking in, also taking into account player/channel mutes. I tested this with all of the supported chat plugins on 1.19.2. If no chat hook is enabled, typing indicators will be sent for the global channel.

Typing indicators are sent as infrequently as possible to maximize performance: if the bot isn't already typing in a channel, it'll start typing immediately, otherwise it'll make sure at least 5 seconds have passed before sending another indicator. Overall, it only takes a couple milliseconds to process each chat preview event even with chat hooks.

Closes #1412.

https://user-images.githubusercontent.com/64972126/195221730-69139306-7212-4499-9437-5fd7175507fc.mp4

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -1147,6 +1147,7 @@ public class DiscordSRV extends JavaPlugin {
 
         // register events
         new PlayerBanListener();
+        new PlayerChatPreviewListener();
         new PlayerDeathListener();
         new PlayerJoinLeaveListener();
         try {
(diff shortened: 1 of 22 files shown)


Vankka:
Chat previewing is being removed in 1.19.3 https://feedback.minecraft.net/hc/en-us/articles/9994475500301-Minecraft-Java-Edition-Snapshot-22w42a

PR #1479, opened 2022-11-23
Switch builds to Gradle with Kotlin DSL

Scarsz:
# To-do:
- [x] Add back the deleted tests
- [x] Get deployment to the repository working
- [x] Add license header plugin
- [x] Add build hash info plugin
- [x] Tidy up final jar name
- [x] Update github actions to use gradle instead of maven


build.gradle.kts
```diff
@@ -0,0 +1,161 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    kotlin("jvm") version "1.6.20"
```
Vankka, line 4: This should be unneccesary, and might even be including kotlin's stdlib in the runtime classpath

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+val targetJavaVersion = 1.8
+
+java {
+    val javaVersion = JavaVersion.toVersion(targetJavaVersion)
+    sourceCompatibility = javaVersion
+    targetCompatibility = javaVersion
+}
+
+tasks {
+    processResources {
```
Vankka, line 20: These tasks don't normally need to be wrapped in `tasks {}`, unless that's something specific to kotlin script for some reason

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+        exclude("META-INF/*.DSA")
+        exclude("META-INF/*.RSA")
+        exclude("META-INF/maven/**")
+        exclude("META-INF/proguard/**")
+    }
+}
+
+fun configureKotlin(kotlinCompile: Action<KotlinCompile>) {
+    tasks.withType<KotlinCompile>().configureEach(kotlinCompile)
+}
```
Vankka, line 71: DiscordSRV doesn't & shouldn't use kotlin, so this should be removed

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+            includeGroup("me.minecraftauth")
+        }
+    }
+}
+
+dependencies {
+    compileOnly("com.destroystokyo.paper:paper-api:${minecraftVersion}-R0.1-SNAPSHOT") {
+        exclude("commons-lang") // we have our own
+    }
+    implementation("net.dv8tion:JDA:4.4.0_352.fix-2") {
```
Vankka, line 93: This should use the `api` configuration

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+        exclude("opus-java") // we don't use voice features
+    }
+    implementation("github.scarsz:configuralize:1.3.2")
+    implementation("org.bstats:bstats-bukkit:2.2.1")
+
+    // logging
+    api("me.scarsz:jdaappender:1.0.3")
```
Vankka, line 100: Does this need to use the `api` configuration?

src/test/java/github/scarsz/discordsrv/CorrectCommonsLangImportTest.java
```diff
@@ -1,53 +0,0 @@
...
-
-import java.io.File;
-import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-
-/**
- * <p>Makes sure that no source files are importing org.apache.commons.lang.*</p>
- * <p>Instead, the import org.apache.commons.lang3.* should be used</p>
- */
-public class CorrectCommonsLangImportTest {
```
Vankka, line 37: The tests can stay as is, just need to add junit as a dependency,

https://docs.gradle.org/current/userguide/java_testing.html#using_junit5
https://docs.gradle.org/current/userguide/java_testing.html#executing_legacy_tests_with_junit_vintage (or the tests could be upgraded to junit 5)

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+    maven("https://papermc.io/repo/repository/maven-public/")
+    maven("https://oss.sonatype.org/content/repositories/central")
+    maven("https://oss.sonatype.org/content/repositories/snapshots")
+    maven("https://nexus.scarsz.me/content/groups/public/") {
+        content {
+            includeGroup("github.scarsz")
+            includeGroup("me.scarsz")
+            includeGroup("me.minecraftauth")
+        }
+    }
```
Vankka, line 86: Aren't like over half of the plugin integrations coming from this repo? They should be included in the `content {}`-block or it should be removed

build.gradle.kts
```diff
@@ -0,0 +1,161 @@
...
+fun configureKotlin(kotlinCompile: Action<KotlinCompile>) {
+    tasks.withType<KotlinCompile>().configureEach(kotlinCompile)
+}
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+    maven("https://hub.spigotmc.org/nexus/content/repositories/snapshots/")
+    maven("https://papermc.io/repo/repository/maven-public/")
+    maven("https://oss.sonatype.org/content/repositories/central")
```
Vankka, line 78: This looks like a mirror of central, which is already included above

Vankka (changes requested):
~~`gradlew`, `gradlew.bat` and `gradle-wrapper.jar` ([+action](https://github.com/DiscordSRV/Ascension/blob/88fd6f4d5da2745d08324f9225b059902c948a72/.github/workflows/gradle-wrapper-validation.yaml) see [extra details](https://github.com/gradle/wrapper-validation-action)) files are missing. Running the `wrapper` task with gradle should fix that~~

build.gradle.kts
Scarsz: > that's something specific to kotlin script for some reason

It is, as far as I can tell, could be wrong

build.gradle.kts
Scarsz: Copied that over from Ascension's build script but you're right

build.gradle.kts
Scarsz: Not really, no, can be downgraded. Left it in because other plugins might want to customize DiscordSRV's JDA appender config but I guess that's a very small use case

build.gradle.kts
```diff
@@ -0,0 +1,150 @@
...
+repositories {
+    mavenLocal()
+    mavenCentral()
+    maven("https://hub.spigotmc.org/nexus/content/repositories/snapshots/")
+    maven("https://papermc.io/repo/repository/maven-public/")
+    maven("https://oss.sonatype.org/content/repositories/snapshots")
+    maven("https://nexus.scarsz.me/content/groups/public/")
+}
+
+dependencies {
```
Vankka, line 76: On the topic of dependency configurations,

- `adventure-api` needs to be `api` since Components are exposed in events, MessageUtil
  - adventure serializers, minimessage and mcdiscordreserializer should be `api`
- `configuralize` should probably be `api`, somebody is probably getting something from there
- `org.jetbrains:annotations` should probably be `compileOnlyApi` to avoid people getting a bunch of errors in decompiled files

Not sure where we should draw the line for the rest of the (internal-ish) dependencies, or if we should just make most dependencies api. I know that [Staff-Chat](https://github.com/DiscordSRV/Staff-Chat/blob/92d814b5835270ff54d7c4947949b9468d7145df/src/main/java/com/rezzedup/discordsrv/staffchat/MessageProcessor.java#L33) uses emoji-java

build.gradle.kts
Scarsz: If I remove the Kotlin plugin, Gradle can't parse the build script at all

It does seem like it's including the Kotlin runtime in the jar though, might need to be excluded?

Vankka: Probably need to add `java-library`, that plugin is likely be providing that transitively

Scarsz: What would we do without you

Vankka (approved):

Vankka:
Probably should change branch if this is going to be v2

Scarsz:
> Probably should change branch if this is going to be v2

No, the switch to Gradle can be a part of v1 and we can then work on v2 in IntelliJ on a separate branch without having to juggle build tools

PR #1481, opened 2022-12-12
Fix errors with javadoc encoding and 'expand' unescaping config/*.yml…

DrakoTrogdor:
Fix errors with javadoc encoding and 'expand' unescaping config/*.yml backslashes

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -58,6 +58,7 @@ tasks {
         withSourcesJar()
     }
     javadoc {
+        options.encoding = "UTF-8"
         options {
             (this as CoreJavadocOptions).addStringOption("Xdoclint:none", "-quiet")
         }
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -58,14 +58,15 @@ tasks {
...
+        options.encoding = "UTF-8"
         options {
             (this as CoreJavadocOptions).addStringOption("Xdoclint:none", "-quiet")
         }
     }
 
     processResources {
         filteringCharset = "UTF-8"
-        filesMatching(listOf("plugin.yml", "config/*.yml")) {
+        filesMatching(listOf("plugin.yml")) {
```
Vankka, line 69: This replacement still needs to occur. Didn't know this expansion screwed with escapes so thanks for that

PR #1484, opened 2022-12-25
Update LunaChat Hook

2b2tjb:
## What did I do?
- Add LunaChatBukkitPostJapanizeEvent Hook
- Add LunaChatConvertedMessageFormat to the messages.yml
- Corrected Japanese translation
### Why?
Many people using LunaChat use the Japanese conversion feature (called Japanize in LunaChat) as well as the channel chat function. However, when using the channel chat function, the Japanese conversion string is not sent to Discord. Therefore, I wrote code so that the Japanese conversion string can also be sent to Discord.

And some Japanese translations were strange, so I fixed them.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -23,6 +23,7 @@
 package github.scarsz.discordsrv;
 
 import alexh.weak.Dynamic;
+import com.github.ucchyocean.lc3.bukkit.event.LunaChatBukkitPostJapanizeEvent;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.gson.Gson;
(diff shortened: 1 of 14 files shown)


src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -1790,16 +1791,32 @@ public void processChatMessage(Player player, Component message, String channel,
...
+                    ? LangUtil.Message.CHAT_TO_DISCORD.toString()
+                    : LangUtil.Message.CHAT_TO_DISCORD_NO_PRIMARY_GROUP.toString())
+                    .replaceAll("%time%|%date%", TimeUtil.timeStamp())
+                    .replace("%channelname%", channel != null ? channel.substring(0, 1).toUpperCase() + channel.substring(1) : "")
+                    .replace("%primarygroup%", userPrimaryGroup)
+                    .replace("%username%", username)
+                    .replace("%usernamenoescapes%", MessageUtil.strip(player.getName()))
+                    .replace("%world%", player.getWorld().getName())
+                    .replace("%worldalias%", MessageUtil.strip(MultiverseCoreHook.getWorldAlias(player.getWorld().getName())));
+        }
```
Vankka, line 1819: Would like to avoid referencing the LunaChat class here as it may cause issues if LunaChat isn't being used. (Probably by just passing the message formats in as parameters to the `processChatMessage` method)

The code in the if and else blocks are nearly identical, we should avoid copying the code, by moving the replace calls outside of the if, example:
```java
if (... lunachat check...) {
  discordMessage = ...lunachat format...;
} else {
  discordMessage = ...normal format...;
}

discordMessage = discordMessage
  .replace("...", ...)
  .replace("...", ...)
...
```

Vankka (commented):
Am I understanding this right, that there would be a secondary message containing the native text and another with the roman version?
eg.
```
Vankka > konnichiwa
[JP] こんいちわ
```

Thanks for the PR!

2b2tjb:
Your understanding is right.

It is certainly better that way.
I’ll update.

GreemDev (approved):

Scarsz:
I'd prefer having this as a separate config file. The main config file should be options for DiscordSRV specifically.

A `lunachat.yml` file could be stored in the resources and only saved if that plugin is detected

src/main/resources/messages/en.yml
```diff
@@ -97,6 +97,37 @@ MinecraftChatToDiscordMessageFormatNoPrimaryGroup: "%displayname% » %message%"
...
+#                  example: Mainland
+# %date%:         current date & time
+#                  example: Sun Jan 1 15:30:45 PDT 2017
+# %channelname%:  the name of the channel that the message was sent in, if the message was sent in a channel at all
+#                  example: Global
+# PlaceholderAPI placeholders are also supported
+
+LunaChatConvertedMessageFormat: "`[JP] %message%`"
+LunaChatConvertedMessageFormatNoPrimaryGroup: "`[JP] %message%`"
+
```
Scarsz, line 130: Move to separate `lunachat.yml` file

src/main/resources/messages/en.yml
```diff
@@ -97,6 +97,37 @@ MinecraftChatToDiscordMessageFormatNoPrimaryGroup: "%displayname% » %message%"
...
+# %world%:        name of world player is in
+#                  example: world
+# %worldalias%:   alias of world player is in via Multiverse-Core
+#                  example: Mainland
+# %date%:         current date & time
+#                  example: Sun Jan 1 15:30:45 PDT 2017
+# %channelname%:  the name of the channel that the message was sent in, if the message was sent in a channel at all
+#                  example: Global
+# PlaceholderAPI placeholders are also supported
+
```
Scarsz, line 127: No trailing newline after comments

Scarsz (changes requested):

2b2tjb:
I apologize for not being able to reply for several months due to being busy.

While I think that the idea of separating the config files is great, the current `messages.yml` file contains a mixture of Dynmap and channel chat settings, and it feels like the consistency of the files would be lost if only the lunachat config is separated.

I think it would be easier for users to either create a new config file for the hook plugin (`hooks.yml`) or integrate it into the `messages.yml` file. What do you think?

Scarsz:
Closing due to inactivity, feel free to update & re-open

granny:
@2b2tjb I like the idea of having a `hooks.yml` config that includes all the hook-related options. I've gone ahead and [created an issue for it](https://github.com/DiscordSRV/DiscordSRV/issues/1715) to make sure it's not forgotten. It should be done as it's own PR.

Issue #1491, opened 2023-01-11
Exempt permission to skip Discord linking


LKits:
### Why?

The main reason to add a permission to exempt for Discord linking is avoiding mandatory linking for certain groups, players.

### What and How?

Not check required linking for players with certain permission, for example: discordsrv.linking.exempt

### Are there alternatives?

No.

### Checks

- [X] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

PR #1510, opened 2023-03-03
Added linking bypass perm

supercoolspy:
based on #1491

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -1320,6 +1320,12 @@ public class DiscordSRV extends JavaPlugin {
             DiscordSRV.warning("/discord command is being handled by plugin other than DiscordSRV. You must use /discordsrv instead.");
         }
 
+        if (config().getBoolean("Require linked account to play.Bypass permission")) {
+            if (!Objects.equals(config.getString("Require linked account to play.Listener event"), "PlayerJoinEvent") || !Objects.equals(config().getString("Require linked account to play.Listener priority"), "NORMAL")) {
+                DiscordSRV.warning("Require linked account to play.Bypass permission is enabled but the listener event is not PlayerJoinEvent with priority of NORMAL, HIGH, or HIGHEST. This will cause issues with the bypass permission.");
+            }
+        }
+
         alertListener = new AlertListener();
         jda.addEventListener(alertListener);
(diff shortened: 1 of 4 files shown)


Scarsz:
Seems unnecessary to have a toggle for the permission, that's the permission plugin's job

That said, the reason this has never been implemented is because the link requirement check happens before the player has even fully joined the server... this means that (most?) permissions plugins don't have the player's permissions loaded just yet.

Have you tested this? Which permissions plugin?

supercoolspy:
> Seems unnecessary to have a toggle for the permission, that's the permission plugin's job
> 
> That said, the reason this has never been implemented is because the link requirement check happens before the player has even fully joined the server... this means that (most?) permissions plugins don't have the player's permissions loaded just yet.
> 
> Have you tested this? Which permissions plugin?

Haven't tested it yet, I've seen some perm loading go fast enough but it just depends what you have loaded. I'll try it right now with a server with 30ish plugins

supercoolspy:
Yeah, doesn't load fast enough

underscore11code:
Did some poking around, currently DSRV defaults to listening to `AsyncPlayerPreLogin` at `LOWEST`, which is before LuckPerms loads at LOW. By `PlayerLogin` at `LOW`, LP will kick players if their data isn't loaded, so any priority after `LOW` on non-async should guarantee permission data, and I believe even anything after `LOW` on async should be fine too (I think LP blocks the async event?). So, switching the event priority to `NORMAL` should be sufficient to make this work.

Obviously this only applies to LP, but given it's got the vast majority of the marketshare, I figure that's probably all that really matters.

gerolndnr:
> Seems unnecessary to have a toggle for the permission, that's the permission plugin's job
> 
> That said, the reason this has never been implemented is because the link requirement check happens before the player has even fully joined the server... this means that (most?) permissions plugins don't have the player's permissions loaded just yet.
> 
> Have you tested this? Which permissions plugin?

If I understand this correctly, this feature would surely be appreciated by many server owners. I looked for a way to enforce discord linking at a later point than first join to make the first join as user-friendly as possible. The plan was to give beginner ranks the bypass permission and as soon as they got used to the server concept and ranked up, they would need the discord linking. 
Is this possible with existing features? I didn't found anything in the docs on the `linking.yml` page.

Scarsz:
Closing this due to inactivity, feel free to re-open if a better solution to this problem is proposed

Issue #1494, opened 2023-01-22
Role colors and names from roleid with placeholders.


LilSadPanda:
### Why?

It would be an alternative to the following placeholder to directly use a role instead of top role etc.

```
%discordsrv_user_top_role_name%
%discordsrv_user_top_role_color_hex%
%discordsrv_user_top_role_color_code%
```

### What and How?

Implement new placeholders that can return a color or name from given role ID for example

```
%discordsrv_roleid_name% = Cool Role Name
%discordsrv_roleid_color_hex% = #FFFFFF
%discordsrv_roleid_color_code% = &F&F&F&F&F&F
```

### Are there alternatives?

Not that I know of :D.

### Checks

- [X] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

Scarsz:
> I think thats a really cool idea, the only thing is would every time the placeholder field is updated be a request to the discord api. Which depending on server sizes and could hit the 50 request/s limit if all scoreboards are updated at the same time and there isn’t a cache

Roles in the guild are cached by the bot internally, there would be minimal performance overhead

NullTxt:
> > I think thats a really cool idea, the only thing is would every time the placeholder field is updated be a request to the discord api. Which depending on server sizes and could hit the 50 request/s limit if all scoreboards are updated at the same time and there isn’t a cache
> 
> Roles in the guild are cached by the bot internally, there would be minimal performance overhead

Alright I'll start working on this at some point using the relational placeholders (makes it so they start with rel_discordsrv)

granny:
implemented in the latest snapshot builds: https://snapshot.discordsrv.com

PR #1511, opened 2023-03-04
Placeholders for all roles

supercoolspy:
Adds placeholders for all roles based on #1494 placeholders would end up being:
```
%discordsrv_role_roleid_name% = Cool Role Name
%discordsrv_role_roleid_color_hex% = #FFFFFF
%discordsrv_role_roleid_color_code% = &F&F&F&F&F&F (Converted to color and would not show up as text)
```


supercoolspy:
@Vankka can you re-run the build? I update from dev

src/main/java/github/scarsz/discordsrv/hooks/PlaceholderAPIExpansion.java
```diff
@@ -113,6 +113,26 @@ public class PlaceholderAPIExpansion extends PlaceholderExpansion {
                 return String.valueOf(accountLinkManager.getLinkedAccountCount());
         }
 
+        if(identifier.startsWith("role_")) {
```
Vankka, line 116:
```suggestion
        if (identifier.startsWith("role_")) {
```
Styling nitpick

Vankka (changes requested):
Thank you for your PR and sorry about the late response

I would prefer if the roleId logic was shared for these placeholders, the regular expression could also use a group to catch the role id instead of removing everything besides numbers. Example of what I mean:
```java
private static final Pattern SPECIFIC_ROLE_PATTERN = Pattern.compile("role_(\\d+)_(\\w+)");

Matcher rolePlaceholderMatcher = SPECIFIC_ROLE_PATTERN.matcher(identifier);
if (rolePlaceholderMatcher.matches()) {
    String roleId = rolePlaceholderMatcher.group(1);
    Role role = DiscordUtil.getRole(roleId);
    String subPlaceholder = rolePlaceholderMatcher.group(2);
    
    if (subPlaceholder.equals("name")) {
        ...
    }
}
```

Dinty1 (approved):
Tested this locally and seems to work perfectly. Only potential barrier to merging is what Vankka said earlier.

Dinty1 (approved):
LGTM

granny (approved):

PR #1522, opened 2023-04-03
Folia support

kugge:
Add [Folia](https://github.com/PaperMC/Folia) support.

* Lag checker and watchdog don't work on Folia so they're disabled by default
* Use reflection to avoir changing API (if you want me to convert it to Folia API ping me)
* Need more tests

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordBanListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordBanListener.java
@@ -22,6 +22,7 @@ package github.scarsz.discordsrv.listeners;
 
 import github.scarsz.discordsrv.Debug;
 import github.scarsz.discordsrv.DiscordSRV;
+import github.scarsz.discordsrv.util.SchedulerUtil;
 import github.scarsz.discordsrv.util.LangUtil;
 import net.dv8tion.jda.api.events.guild.GuildBanEvent;
 import net.dv8tion.jda.api.events.guild.GuildUnbanEvent;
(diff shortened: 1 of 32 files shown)


Vankka (changes requested):
I left a few in-line comments/style fixes

kugge:
Thank you for your review. I force pushed the requested changes. I believe that more tests are needed, though I haven't encountered any problem yet.

Vankka:
Kicking players does appear to work on the global scheduler but it should probably occur on the Player's own EntityScheduler (Entity#getScheduler). The other scheduler calls look to be using the appropriate scheduler(s)

Otherwise I'm fine with merging this

kugge:
> Kicking players does appear to work on the global scheduler but it should probably occur on the Player's own EntityScheduler (Entity#getScheduler). The other scheduler calls look to be using the appropriate scheduler(s)
> 
> Otherwise I'm fine with merging this

Done, changes were force-pushed

src/main/java/github/scarsz/discordsrv/listeners/DiscordBanListener.java
```diff
@@ -58,7 +59,7 @@ public void onGuildBan(GuildBanEvent event) {
         if (offlinePlayer.isOnline()) {
             // also kick them because adding them to the BanList isn't enough
             Player player = offlinePlayer.getPlayer();
-            if (player != null) Bukkit.getScheduler().runTask(DiscordSRV.getPlugin(), () -> player.kickPlayer(reason));
+            if (player != null) SchedulerUtil.runTask(DiscordSRV.getPlugin(), () -> player.kickPlayer(reason));
```
Vankka, line 62: This needs to use the new method as well

kugge: Done. Don't hesitate to ping me if I've forgotten something or if tests fails.

kugge:
Added Scheduler caching
Will consider adding Method Handlers

Vankka (approved):

PR #1523, opened 2023-04-07
Fix voice

supercoolspy:


supercoolspy:
Oops gotta remove all the extra commits ill do that when I get home

Vankka:
DiscordSRV doesn't use audio, and excludes the opus binaries required to do so

PR #1530, opened 2023-04-23
Moved CodeGenerated value

okunamayanad:
Moved CodeGenerated under Discord to being under Minecraft


James-Bennett-295 (changes requested):
You should also make this change in [the other messages.yml files](https://github.com/DiscordSRV/DiscordSRV/tree/develop/src/main/resources/messages).

James-Bennett-295 (approved):

Vankka (approved):

PR #1534, opened 2023-05-22
build: update dynmap to support newer versions

ANutley:


Vankka:
I don't see how these changes would fix the linked issue

Scarsz:
> I don't see how these changes would fix the linked issue

[In the DynmapHook](https://github.com/DiscordSRV/DiscordSRV/blob/767828469573c2f0fa3ed5a44feb0b2f15f2c02d/src/main/java/github/scarsz/discordsrv/hooks/DynmapHook.java#LL52C65-L52C65), presumably, the class inheritance/hierarchy may have changed, though that would probably require a code change as well, no?

@ANutley have you tested this?

ANutley:
Upon further investigation this issue was caused by another plugin shading the dynmap api, not the version of the api that we use.

PR #1567, opened 2023-07-19
Fix console stripping

James-Bennett-295:
Fix the `aggressiveStripPattern` so that it matches all ANSI sequences, and doesn't cause numbers to be removed.


Vankka (approved):
Tested Paper 1.8, 1.19 and 1.20, seems fine

James-Bennett-295:
> @James-Bennett-295 The numbers after the color code are still removed. ex) `&a10%` → `%`

A different regex is used for those type of colour codes: `(?<!<@)[&§\u007F](?i)[0-9a-fklmnorx]`

The regex I changed and the regex above will not cause that.

PR #1569, opened 2023-07-20
Fix texture string extraction

James-Bennett-295:
Currently the regex pattern for extracting texture strings is `https?://.+(?<texture>\\w{64})\"`, which is usually ok as the JSON contains
```
    "SKIN" : {
      "url" : "http://textures.minecraft.net/texture/..."
    },
    "CAPE" : {
      "url" : "http://textures.minecraft.net/texture/..."
   }
```
so the SKIN url's texture will be extracted as it is first, however a user in a support ticket has a player who's texture string is only 63 characters, so was not matched by the regex causing the CAPE url's texture to be extracted instead. I have changed the regex to fix this.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/util/NMSUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/NMSUtil.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 
 public class NMSUtil {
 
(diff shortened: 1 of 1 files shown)


James-Bennett-295:
The reason for the texture hash being below 64 characters seems to be due to mojang not zero-padding them.

Scarsz:
I don't like how lenient that expression is now. I'd settle for clamping it as `{60,64}`

PR #1573, opened 2023-07-22
Don't count console and chat channel IDs as equal if both are unset

James-Bennett-295:
Makes it so debug reports won't fail the `CONSOLE_AND_CHAT_SAME_CHANNEL` test if the console and chat channel IDs are both unset.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/util/DebugUtil.java
+++ b/src/main/java/github/scarsz/discordsrv/util/DebugUtil.java
@@ -295,8 +295,8 @@ public class DebugUtil {
         }
 
         String consoleChannelId = DiscordSRV.config().getString("DiscordConsoleChannelId");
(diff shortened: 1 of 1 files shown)


src/main/java/github/scarsz/discordsrv/util/DebugUtil.java
```diff
@@ -295,7 +295,7 @@ private static String getDebugInformation() {
         }
 
         String consoleChannelId = DiscordSRV.config().getString("DiscordConsoleChannelId");
-        if (DiscordSRV.getPlugin().getChannels().values().stream().filter(Objects::nonNull)
+        if (!consoleChannelId.matches("^0*$") && DiscordSRV.getPlugin().getChannels().values().stream().filter(Objects::nonNull)
```
Vankka, line 298: This will cause a error if `consoleChannelId` is null

Vankka (changes requested):

Vankka (approved):

PR #1607, opened 2023-11-12
Correction of Japanese

siyukatu:
"に返信する %name%" is incorrect.
In English, for example, this would be something like "%name% Reply to".
The order of the sentences is reversed.
Therefore, i have made the change to "%name% への返信".


granny:
Thank you for your contribution! :)

PR #1624, opened 2024-01-11
Switch to a more stable avatar service

granny:


Scarsz (changes requested):
Considering Crafatar has made it clear that they don't want traffic from Discord *ever*, probably not necessary to leave the previous URL as a comment

PR #1647, opened 2024-02-25
Improve Chinese translation

SamLam140330:
Fix some incorrect translations and convert some simplified Chinese to traditional Chinese as zh means traditional Chinese.


granny (approved):
Thank you for your PR!

PR #1648, opened 2024-02-26
Update Chatty api to 3.0.0-SNAPSHOT

sub-kek:
Fix DiscordSRV does not work with Chatty version 3.0.0-SNAPSHOT. This will break hook Chatty 2.19.14 or below.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -262,7 +262,7 @@ dependencies {
     ///
     
     // chat hooks
-    compileOnly("ru.mrbrikster:chatty-api:2.18.2")
+    compileOnly("ru.brikster:chatty-api:3.0.0-SNAPSHOT") // published to maven local
     compileOnly("br.com.finalcraft:fancychat:1.0.2")
     compileOnly("com.dthielke.herochat:Herochat:5.6.5")
     compileOnly("br.com.devpaulo:legendchat:1.1.5")
--- a/src/main/java/github/scarsz/discordsrv/hooks/chat/ChattyChatHook.java
+++ b/src/main/java/github/scarsz/discordsrv/hooks/chat/ChattyChatHook.java
@@ -31,9 +31,9 @@ import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.plugin.Plugin;
-import ru.mrbrikster.chatty.api.ChattyApi;
-import ru.mrbrikster.chatty.api.chats.Chat;
-import ru.mrbrikster.chatty.api.events.ChattyMessageEvent;
+import ru.brikster.chatty.api.ChattyApi;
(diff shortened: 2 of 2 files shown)


granny (changes requested):
Blocking this PR until builds of [Chatty](https://github.com/Brikster/Chatty) v3.0.0 become public

sub-kek:
This PR is for the future.

granny:
> This PR is for the future.

Right. I made sure to make a note of that so it doesn't get merged haphazardly.

Scarsz (changes requested):
Not a fan of breaking the compatibility with Chatty 2.x, there can be two versions of the hook that can be dynamically chosen

Boddebon:
Hello, is it possible to make it so that users can enable this at their own risk, because some servers have already switched to Chatty V3, but they disable chats in Discord because of this problem?

Dinty1:
If Scarsz’ comment is addressed it could probably be merged

PadowYT2:
> Hello, is it possible to make it so that users can enable this at their own risk, because some servers have already switched to Chatty V3, but they disable chats in Discord because of this problem?

I could help you build it yourself, or straight up send you the .jar if you want. DM me on Discord (IGN - `padowyt2`)

granny:
> > Hello, is it possible to make it so that users can enable this at their own risk, because some servers have already switched to Chatty V3, but they disable chats in Discord because of this problem?
> 
> I could help you build it yourself, or straight up send you the .jar if you want. DM me on Discord (IGN - `padowyt2`)

I do not recommend anybody do this at all. This is dangerous and you have no guarantee that the person sending you the file didn't make any additional modifications to it.

sub-kek:
Maybe I'll make a second hook for Chatty V3 just without the dependencies.

granny:
> Maybe I'll make a second hook for Chatty V3 just without the dependencies.

@sub-kek Why not depend on both v2 and v3 api's and use the one that's available at runtime? That's what scarsz suggested.

sub-kek:
> > Maybe I'll make a second hook for Chatty V3 just without the dependencies.
> 
> 
> 
> @sub-kek Why not depend on both v2 and v3 api's and use the one that's available at runtime? That's what scarsz suggested.

The v3 api is not published anywhere yet this may cause problems with the build.

granny:
> > > Maybe I'll make a second hook for Chatty V3 just without the dependencies.
> > 
> > 
> > @sub-kek Why not depend on both v2 and v3 api's and use the one that's available at runtime? That's what scarsz suggested.
> 
> The v3 api is not published anywhere yet this may cause problems with the build.

We can publish it on scarsz' nexus repo so that's not a problem.

Boddebon:
Hello, do you have any information?

sub-kek:
> We can publish it on scarsz' nexus repo so that's not a problem.

I think it should be done.  Some people ask how to build it.

Issue #1404, opened 2022-05-17
Apply filter to attachments


JinjaFormbi:
### Why?

as seen in issue  #1403 
"Possible to config certain messages to default ones / remove long discord Picture image link messages inside Minecraft"
Images get send as long text messages. 

### What and How?

Adding uploaded files to the [DiscordChatChannelDiscordFilters](https://github.com/DiscordSRV/DiscordSRV/blob/master/src/main/resources/config/en.yml#L255) 

### Are there alternatives?

Make a specific function that can filter discord files

### Checks

- [X] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

### Anything else

thanks to chrissamael for the specification of the problem

JinjaFormbi:
Currently if the solution in https://github.com/DiscordSRV/DiscordSRV/issues/1403 is applied the link will be send regardless if the image is uploaded by the poster. the filter dies apply if the poster sends the link to the image. 
Both variants of posting are indistinguishable for a user on the discord side on the first glance.

GreenGinkgo:
This appears to still be a problem. I have applied a regex filter to completely wipe out any line that contains "http" and yet file URLs are still sent to chat. My hunch is that it might be due to the text being sent in a separate "message" which is not the text of the message a Discord user sends. It might depend on HOW the link is applied - my test was to copy an image and paste it while composing a Discord message, which triggers some kind of dynamic upload. See screenshots
![image](https://user-images.githubusercontent.com/5168056/207789385-bc58771f-0526-4741-966c-86467b06c890.png)
![image](https://user-images.githubusercontent.com/5168056/207789403-326a121b-6c3e-4dd1-8e89-f2921f8b1198.png)
![image](https://user-images.githubusercontent.com/5168056/207789407-80adde97-f4e0-4b2a-8647-baf5984abc30.png)

granny:
Should be fixed on latest snapshot builds. https://snapshot.discordsrv.com/

PR #1658, opened 2024-03-27
Apply Discord regexes to message addon URLs

James-Bennett-295:
This means images being sent in Discord will have their URL be filtered, like would happen if the URL of the image was sent.


James-Bennett-295:
This would close https://github.com/DiscordSRV/DiscordSRV/issues/1404

PR #1666, opened 2024-04-11
Add boost status placeholder

### LINKED ISSUE #1665: Add the placeholder to know the status of the player if they are a booster
### Why?

In this way it is easy to manage bossters with external plugin conditions

### What and How?

Using the placeholder on a player should return true or false, true if the player is a booster

INPUT: %discordsrv_user_bosster%
OUTPUT: true/no

### Are there alternatives?

Exists only for permissions in luckperms but this cannot be handled with conditions

### Checks

### Anything else

_No response_

thewilloftheshadow:
Closes #1665 

I haven't tested this yet but it seems simple enough, will test tonight


thewilloftheshadow:
Shush I definitely remembered that

PR #1668, opened 2024-04-13
AvatarUrl changes

James-Bennett-295:
This removes the need for people to change the AvatarUrl if there is ever an issue with the default or if they're using offline mode.

The code already chooses an appropriate default value when the URL is empty.

If you are ok with the EN config change then I will do it for the other languages.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -2094,7 +2094,7 @@ public class DiscordSRV extends JavaPlugin {
             avatarUrl = !offline ? defaultUrl : offlineUrl;
             DiscordSRV.config().setRuntimeValue("AvatarUrl", avatarUrl);
 
-            DiscordSRV.error("Your AvatarUrl config option uses crafatar.com which no longer supports Discord so an alternative provider will be used.");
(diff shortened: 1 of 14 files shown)


src/main/java/github/scarsz/discordsrv/DiscordSRV.java
```diff
@@ -2090,20 +2090,17 @@ private static String constructAvatarUrl(String username, UUID uuid, String text
...
+            DiscordSRV.config().setRuntimeValue("AvatarUrl", avatarUrl);
 
-            if (!offlineUuidAvatarUrlNagged) {
-                DiscordSRV.error("Your AvatarUrl config option does not contain the {username} placeholder even though this server is using offline UUIDs.");
-                DiscordSRV.error(offlineUrl + " will be used because the default value does not support offline mode servers");
-                DiscordSRV.error("You should set your AvatarUrl (in config.yml) to " + offlineUrl + " (or another url that supports usernames) "
-                        + (defaultValue ? "to get rid of this error" : "to get avatars to work."));
-                offlineUuidAvatarUrlNagged = true;
-            }
+            DiscordSRV.error("Your AvatarUrl config option uses crafatar.com which no longer supports Discord so an alternative provider will be used.");
```
Scarsz, line 2097:
```suggestion
            DiscordSRV.error("Your AvatarUrl config option uses crafatar.com, which no longer allows usage with Discord. An alternative provider will be used.");
```

Scarsz (changes requested):

PR #1674, opened 2024-05-13
Update fr.yml

PtitLegume:
correcting a translation error


granny (approved):
Thank you for your contribution! :)

PR #1680, opened 2024-05-22
Fixed Blacklist/Whitelist not being resepected

### LINKED ISSUE #1557: %discordsrv_user_top_role_color_hex% not ignored when role is blacklisted from DiscordChatChannelRolesSelection config.
### Expected behavior

%discordsrv_user_top_role_color_hex% should not grab the top role color of a users role that's blacklisted in the setting DiscordChatChannelRolesSelection.

### Actual behavior

%discordsrv_user_top_role_color_hex% not ignored when role is blacklisted from DiscordChatChannelRolesSelection config.
The placeholder still grabs the top role of a user who's top role is blacklisted in the config setting DiscordChatChannelRolesSelection.

### Steps to reproduce

Setup roles you wish to blacklist in the setting DiscordChatChannelRolesSelection and then use the placeholder %discordsrv_user_top_role_color_hex% anywhere you'd like and you'll notice it still grabs the role color of the role blacklisted.

### Server software and version

git-Purpur-2005 (MC: 1.20.1)*

### Checks

- [X] I am __not__ using an outdated version of DiscordSRV.
- [X] I asked in [DiscordSRV's Discord server](https://discord.discordsrv.com) to see whether this issue is in fact a bug that needs to be fixed.

### Anything else

_No response_

AwesomeBFM:
fix #1557 

Added getTopSelectedRole() method to DiscordSRV.java. Method calls getSelectedRoles() of the same scope and attempts to get the highest role the user has that is present in the selected roles.

Changes onRequest in PlaceholderAPIExpansion.java to use this new method rather than just fetching their highest role and ignoring selectedRoles. Additionally, removed logic getting a list of selected rules as it is not needed here.


Dinty1 (changes requested):
After talking about this a bit more it’d be preferable if these were split out into different placeholders (ie top_selected_role)

AwesomeBFM:
Sounds good

granny (approved):
Thank you for the PR! :)

PR #1704, opened 2024-06-26
Comment that nickname sync is controlled by a permission

J-295:
If no changes are wanted for the EN config then I will translate it for the other languages.

### DIFF
--- a/src/main/resources/synchronization/en.yml
+++ b/src/main/resources/synchronization/en.yml
@@ -1,4 +1,5 @@
 # Minecraft -> Discord nickname synchronization
+# Controlled by the 'discordsrv.nicknamesync' permission
 #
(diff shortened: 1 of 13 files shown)


src/main/resources/synchronization/en.yml
```diff
@@ -1,4 +1,5 @@
 # Minecraft -> Discord nickname synchronization
+# Controlled by the 'discordsrv.nicknamesync' permission
```
ANutley, line 2:
```suggestion
# Can be controlled per player through the use of the 'discordsrv.nicknamesync' permission
```

ANutley (approved):

src/main/resources/synchronization/en.yml
Dinty1: And say granted by default maube

ANutley (approved):

PR #1739, opened 2024-08-05
Put commit hash in plugin.yml version

Dinty1:
This would mean that we can see how many people are *actually* running the latest snapshot, which might be useful for making a decision about whether the build is stable enough.


granny (dismissed):
Verify that this doesn't cause issues with the config migrator

granny (commented):
> Vankka — 1/24/25, 3:26 AM
snapshots should probably be migrated more explicitly than just hash changed, since the v1 migration isn't that perfect

PR #1755, opened 2024-08-30
Add placeholders for user ID

fres621:
This adds the placeholder `%userid%` for and Discord -> Minecraft messages and replies
## Why?
Ideal use case is with MiniMessage's `<insert>` to let Minecraft players ping Discord users without having to grab their ID from the Discord. The following messages.yml config example showcases the use of these new placeholders:
```yml
DiscordToMinecraftChatMessageFormat: '%reply%[<aqua>Discord</aqua> | %toprolecolor%%toprolealias%<reset>] <insert:"<@%userid%>">%name%</insert> » %message%'
DiscordToMinecraftChatMessageFormatNoRole: '%reply%[<aqua>Discord</aqua>] <insert:"<@%userid%>">%name%</insert> » %message%'
DiscordToMinecraftAllRolesSeparator: ' | '
DiscordToMinecraftMessageReplyFormat: '<insert:"<@%userid%>"><hover:show_text:"%name% » %message%">[REPLY]</hover></insert> '
```
Now if you hold shift and click a Discord user's name, their mention will be inserted in the Minecraft chat.
Other possible use cases would be `<click>` tags to run commands that use the user's ID
If translation for these is needed, I'm able to do English and Spanish

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/DiscordChatListener.java
@@ -392,6 +392,7 @@ public class DiscordChatListener extends ListenerAdapter {
         return input.replace("%channelname%", event.getChannel().getName())
                 .replace("%name%", escape.apply(MessageUtil.strip(event.getMember() != null ? event.getMember().getEffectiveName() : event.getAuthor().getName())))
                 .replace("%username%", escape.apply(MessageUtil.strip(event.getAuthor().getName())))
+                .replace("%id%", event.getAuthor().getId())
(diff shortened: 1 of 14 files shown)


Vankka:
Should be `%userid%` instead

fres621:
> Should be `%userid%` instead

Done

Vankka (approved):

Dinty1 (changes requested):
This should also be documented in messages.yml with the other placeholders. Just copy-pasting the English across the different files is fine.

fres621:
> This should also be documented in messages.yml with the other placeholders. Just copy-pasting the English across the different files is fine.

Done

Dinty1 (approved):

PR #1770, opened 2024-10-18
Migrate ProxyHost option with new default value

J-295:
Currently, updating from the release to the snapshot means the old ProxyHost default value stays in the config, so the plugin sees the value as non-default and tries to use it.

Since the new default is a correction, it is better to convert it than to have the plugin ignore it too.


Vankka (approved):

andrewkm:
Lol I was about to make an issue about this yesterday, but ended up figuring it out and just updating my configs. xD

PR #1787, opened 2025-01-23
Fix NPE when account link manager hasn't been created yet

ANutley:
On large servers with a lot of users constantly joining and leaving, it isn't rare for people to join the Discord server in between the Discord bot starting and the account link manager being created


Dinty1 (approved):
I’d be astounded if this broke something but still going to wait for confirmation that you’ve tested it

Issue #1625, opened 2024-01-11
Permission to block Discord->Minecraft chat messages


twiceineverymoment:
### Why?

There may arise circumstances where a server owner may want a specific group of players to not have their chat synced up with Discord at all. The permission node `discordsrv.chat` allows certain players' messages to be blocked from sending to Discord when set to false, but there is no way to achieve this in the opposite direction.

### What and How?

Add a permission such as `discordsrv.chat.receive`. Players for whom this permission is set to false do not see any of the incoming messages from the Discord chat. Include this permission under `discordsrv.player`.

### Are there alternatives?

I have not found an alternative method that works. It was suggested in the Discord that I create a feature request for this (ticket 16848.)

### Checks

- [X] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

Goinging:
I am having a similar problem when it comes to players /ignore 'ing other players in Minecraft, but those same players still get messages from the players they ignored coming from the discord. My suggestion is to make it be based on linked minecraft - discord with maybe a toggle in the config. A perm would also work for this purpose if it is automatically applied when someone did /ignore or maybe instead of using essentials /ignore it could use a separate discord srv ignore that ignores both minecraft and discord messages from a player.

Issue #1234, opened 2021-08-24
[Enhancement] Adding /discordsrv ignore -DiscordUsername- to commands list


farum12:
Player Xyxy ignores player Zyzy ingame, in MC via some chat plugin.
Hovewer, player Zyzy uses Discord to send messages to the MC chat.
Player Xyxy doesn't have any option to ignore that and wants to avoid messages from Zyzy.

Therefore I would suggest adding /discordsrv ignore [DiscordUsername] as a command, which would remove specific recipient (player Xyxy) from messages which are sent from specific Discord User (User Zyzy).

Thanks!

Dinty1:
and/or we could add an API method to ignore players that chat plugins can use

Kurkulas:
yeah, i need it on my server

sovtec:
Just yes, this feature would be beautiful

exwundee:
I would love to see this feature added back.

Issue #603, opened 2019-12-31
Feature: Ignore command and/or disable chat bridge messages


Weasel-Beans:
Can there be a command for ignoring users across the chat bridge? Like users in minecraft ignoring discord messages, and/or the entire chat-bridge stream?

Scarsz:
Existed at one point in time, was removed due to literally nobody using it and it not able to be implemented well with chat plugin implementations. Also makes the chat inconsistent with players replying to messages from Discord but not seeing the original message.

https://github.com/DiscordSRV/DiscordSRV/commit/135ad689a38cd20cd31c025a7a4c616a0607174b#diff-0ad22d382c20fe65fee64bbce9d8d345

Vankka:
@Scarsz The request was for In-Game Players to be able to ignore specific Discord users

Scarsz:
Point stands that it's not feasible to implement due to the chat plugin implementations not widely supporting such a thing

PR #1789, opened 2025-01-23
Include author in DiscordGuildMessagePreBroadcastEvent

jenbroek:
In lieu of a native `/ignore` feature not coming back (https://github.com/DiscordSRV/DiscordSRV/issues/603), I'm working on an addon to `/ignore` Discord-to-MC messages from specific users, through subscribing to [`DiscordGuildMessagePreBroadcastEvent`](https://github.com/DiscordSRV/DiscordSRV/blob/master/src/main/java/github/scarsz/discordsrv/api/events/DiscordGuildMessagePreBroadcastEvent.java) and altering recipients. I need the author of the Discord message to do so, however currently this requires storing the author from [`DiscordGuildMessagePostProcessEvent`](https://github.com/DiscordSRV/DiscordSRV/blob/master/src/main/java/github/scarsz/discordsrv/api/events/DiscordGuildMessagePostProcessEvent.java) and later grabbing it based on the message. This is feels like a hacky and cumbersome workaround, given that the author is available in the method that fires `DiscordGuildMessagePreBroadcastEvent`: [`‎DiscordSRV::broadcastMessageToMinecraftServer‎`](https://github.com/DiscordSRV/DiscordSRV/blob/master/src/main/java/github/scarsz/discordsrv/DiscordSRV.java#L1876).

Having the author be accessible from `DiscordGuildMessagePreBroadcastEvent` would eliminate the need for a workaround and make it trivial to implement `/ignore` syncing as an addon (possibly closing https://github.com/DiscordSRV/DiscordSRV/issues/1234 and https://github.com/DiscordSRV/DiscordSRV/issues/1625).

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
+++ b/src/main/java/github/scarsz/discordsrv/DiscordSRV.java
@@ -1882,7 +1882,7 @@ public class DiscordSRV extends JavaPlugin {
         if (chatHook == null || channel == null) {
             if (channel != null && !channel.equalsIgnoreCase("global")) return; // don't send messages for non-global channels with no plugin hooks
             DiscordGuildMessagePreBroadcastEvent preBroadcastEvent = api.callEvent(new DiscordGuildMessagePreBroadcastEvent
-                    (channel, message, PlayerUtil.getOnlinePlayers()));
+                    (author, channel, message, PlayerUtil.getOnlinePlayers()));
             message = preBroadcastEvent.getMessage();
             channel = preBroadcastEvent.getChannel();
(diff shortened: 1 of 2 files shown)


Scarsz (changes requested):
Missing an import for the `User` class in the event

jenbroek:
> Missing an import for the `User` class in the event

Whoops, fixed.

Scarsz (approved):

granny (approved):

granny:
Feel free to mention your ignore plugin in the mentioned issue trackers once it becomes public.

PR #1791, opened 2025-02-11
add broadcast event to broadcast command

monckey100:
This event allows users to block the default broadcast event and use alerts.yml to customize the broadcast. This is perfect for scenarios where the user has a plugin that uses the command and therefore no events are fired. This also allows other plugins to interact with the event if they choose to listen to the event.

in your config.yml:
`Experiment_BlockBroadcast: true`
to block the default broadcast

```
  - Trigger: DiscordSRVBroadcastEvent
    Channel: ingame
    Embed:
      Color: "#ffcc00"
      Author:
        ImageUrl: "{botavatarurl}"
        Name: "${message}"
      Timestamp: true
      ```


Vankka (changes requested):
There should be a new command such as `/discord alert <some identifier>` rather than blocking out the broadcast command (the identifier would be the trigger in the alerts.yml, preferably with some prefix to distinguish from class names)

Sorry about the late response to this PR

PR #1792, opened 2025-02-23
Expand DiscordGuildMessagePreBroadcastEvent API

Jsinco:
This PR expands this class' API so that the event is no longer *just* tied to broadcasting and now supports all the external plugins DiscordSRV hooks into.

It's not perfect, and the `recipients` list is unmodifiable in every `ChatHook` except for **TownyChat** and **VentureChat** because modifying the recievers isn't possible without changing the chat channel.

Majority of servers are using external chat plugins and not being able edit who can recieve XYZ message without cancelling the entire beforehand with something like `DiscordGuildMessagePreProcessEvent` limits what we can do with the API. _(e.g. No `/ignore` command capabilities or ignoring a specific player, no turning off DIscordSRV in-game chat for specific players, etc.)_

Some of these `ChatHook`s can also be edited to support modifying the recipient list later on in this PR too.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -180,7 +180,7 @@ repositories {
     mavenLocal()
     mavenCentral()
     maven("https://hub.spigotmc.org/nexus/content/repositories/snapshots/")
-    maven("https://papermc.io/repo/repository/maven-public/")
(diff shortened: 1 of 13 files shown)


TheosisOfficial:
Please can this be merged!

Issue #1793, opened 2025-03-02
Major alerts lag on 1.29.0 and later


Vankka:
### Expected behavior

Alerts for events that are not being used should be unsubscribed

### Actual behavior

They are not, and cpu usage is high on the server thread

### Steps to reproduce

Use alerts at all

### Server software and version

n/a

### Checks

- [x] I am __not__ using an outdated version of DiscordSRV.
- [x] I asked in [DiscordSRV's Discord server](https://discord.discordsrv.com) to see whether this issue is in fact a bug that needs to be fixed.

### Ticket number or message link to where you asked in DiscordSRV's Discord server about this

-

### Anything else

Regression from https://github.com/DiscordSRV/DiscordSRV/pull/1750

bobhenl:
Looking forward to this

MC-MrBirdy:
Is this related to an error that I'm experiencing during startup/reloads:
```[ERROR]: [DiscordSRV] java.lang.NoSuchFieldException: modifiers
[ERROR]: [DiscordSRV]  at java.base/java.lang.Class.getDeclaredField(Class.java:2782)
[ERROR]: [DiscordSRV]  at io.papermc.reflectionrewriter.runtime.AbstractDefaultRulesReflectionProxy.getDeclaredField(AbstractDefaultRulesReflectionProxy.java:90)
[ERROR]: [DiscordSRV]  at io.papermc.paper.pluginremap.reflect.PaperReflectionHolder.getDeclaredField(Unknown Source)
[ERROR]: [DiscordSRV]  at DiscordSRV-1.30.0-SNAPSHOT-d55304f.jar//github.scarsz.discordsrv.modules.alerts.AlertListener.register(AlertListener.java:118)
[ERROR]: [DiscordSRV]  at DiscordSRV-1.30.0-SNAPSHOT-d55304f.jar//github.scarsz.discordsrv.modules.alerts.AlertListener.reloadAlerts(AlertListener.java:184)
[ERROR]: [DiscordSRV]  at DiscordSRV-1.30.0-SNAPSHOT-d55304f.jar//github.scarsz.discordsrv.modules.alerts.AlertListener.<init>(AlertListener.java:101)
[ERROR]: [DiscordSRV]  at DiscordSRV-1.30.0-SNAPSHOT-d55304f.jar//github.scarsz.discordsrv.DiscordSRV.init(DiscordSRV.java:1380)
[...]

PR #1799, opened 2025-04-16
Add support for using fully qualified names in alerts, deprecate using simple names in alerts

Vankka:
There is no way to listen for events without knowing the entire class name. 
The current hack to get around this is to register a listener to every event, but this comes with various problems such as:
- Performance problems (some events trigger *very* frequently)
  + Being unable to unregister from unused events reliably, due to some events sharing their HandlerList (see below attempted fix and resulting lag problems)
- Having to replace the allLists field value to detect new events being added to it
  + This field is final [as of Paper 1.21.5](https://github.com/PaperMC/Paper/commit/f00727c57e564f3a8cb875183a54142feb693db7#diff-4cee0ef308b207a5a696351757900f4ff517340c4a655754486cf9ba45dcd7d2R34)

Fixes GH-1793 (In cases where no simple names are used)  
Fixes GH-1814

Docs PR: https://github.com/DiscordSRV/Documentation/pull/85

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/modules/alerts/AlertListener.java
+++ b/src/main/java/github/scarsz/discordsrv/modules/alerts/AlertListener.java
@@ -212,10 +212,10 @@ public class AlertListener implements Listener, EventListener {
                     Method method = null;
                     Class<?> handlerListClass = eventClass;
                     while (method == null && handlerListClass != null) {
-                        handlerListClass = handlerListClass.getSuperclass();
                         try {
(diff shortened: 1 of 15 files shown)


ANutley:
https://github.com/DiscordSRV/DiscordSRV/blob/e6dbac629758140d0585d85ac626fa34b3d62788/src/main/resources/alerts/en.yml#L50

Would it be worth changing this to something along the lines of "event fully qualified class name", however we may want something simpler that the average server owner would understand?

OakLoaf:
Is this safe to compile and use in production?

PR #1801, opened 2025-04-28
Portuguese Translation

zyypj:
- Added pt.yml, containing messages and comments from en.yml translated to Portuguese.
(Brazilian Portuguese)


thewilloftheshadow (changes requested):
The other folders in the resources folder need to be translated as well in order to add a translation to DiscordSRV.

ANutley:
The [LangUtil](https://github.com/DiscordSRV/DiscordSRV/blob/develop/src/main/java/github/scarsz/discordsrv/util/LangUtil.java) file must also be translated

PR #1807, opened 2025-05-11
Add support for MultiverseCore API v5

Bermiin:
Added support for the new Multiverse Core v5 which introduced API breaking changes, which resulted in breaking DiscordSRV's Minecraft -> Discord message forwarding feature.

Multiverse Core v4 is still supported


benwoo1110:
That’s great!! Thanks! 

~ fellow multiverse dev here

PR #1815, opened 2025-06-19
Detect bedrock users and replace with fallback avatar url

Inpuzah:
Useful for servers that are running geyser/floodgate that allows for cross platform play between Java and bedrock clients on the same server.

Needs more implementation for servers who change the bedrock prefix from "." to something else, however "." is the default configuration.


Vankka:
This wouldn't be very useful unless the server only allows bedrock players to join, since the same username may have a different skin on Bedrock and Java, and only one URL can be specified in the config.  
So far the approach for Floodgate users has been to use the Player UUID (which has the first half as all 0s for Bedrock players) and a skin API that has special handling for those

Vankka:
Closing as I don't think this is useful for the reasons specified above

PR #1837, opened 2025-10-04
only print one error message in topic updater

minneelyyyy:
This fixes what I perceive to be a flaw where the bot will flood the console if it doesn't have the permission to manage channels.

image

It replaces that with the following message
image

This PR also necessarily changed `DiscordUtil.setTextChannelTopic` to throw a `PermissionException` when your bot lacks permission to set the channel settings, instead of printing a generic message for that case. Which, I think is slightly better anyway, but it did require touching code in `DiscordSRV` as well.


granny (approved):
Thank you for the PR! I made some additional changes but it continues to only print one warning message in topic updater instead of spamming permission errors

Issue #1868, opened 2026-04-22
Resolve MiniMessage tags in `MinecraftChatToDiscordMessageFormat` and `MinecraftChatToDiscordMessageFormatNoPrimaryGroup`


OfficialTheosis:
### Why?

Because if I have a LuckPerms prefix that uses MiniMessage tags instead of legacy ones, I don't want to see the actual tags themselves in the Discord channel like this:
<img width="608" height="29" alt="Image" src="https://github.com/user-attachments/assets/58ad17e4-1af8-4736-9086-dc5e9bbbc99d" />

### What and How?

Using `PlainTextComponentSerializer.plainText().serialize(...)` on the messages should be sufficient to remove the MiniMessage tags and just end up with raw text to send to Discord.

### Are there alternatives?

Currently we have to use legacy ampersand characters in things like LuckPerms prefixes and use PlaceholderAPI expansions like [Kyorify](https://api.extendedclip.com/expansions/kyorify/) to convert them into MiniMessage strings for every other plugin that supports MiniMessage.

### Checks

- [x] I have used the search at least once to check if my idea has already been suggested and perhaps already implemented.

PR #1870, opened 2026-05-16
Flatten MiniMessage tags in Minecraft -> Discord chat format

BoredManCodes:
LuckPerms prefixes and PAPI expansions commonly return MiniMessage now (eg. <gradient:red:blue>Owner</gradient>). Because the chat format pattern itself is never reserialized to Discord markdown, those tags appeared in chat as literal text rather than the rendered prefix.

Add MessageUtil.stripMiniMessage which parses input as MiniMessage and serializes via PlainTextComponentSerializer so the tag syntax is removed while the underlying text is preserved. Apply it to the substituted format pattern in DiscordSRV.translateAndSendToDiscord, after PAPI substitution but before %message% is filled in (so the player's own message body is unaffected).

Closes DiscordSRV/DiscordSRV#1868


OfficialTheosis:
Please merge this! 🙏

PR #1881, opened 2026-08-07
Return onPlayerAdvancementDone if advancement message is null

DaMarine:
As per the [Paper docs](https://jd.papermc.io/paper/26.2/org/bukkit/event/player/PlayerAdvancementDoneEvent.html), `message()` is a `@Nullable` Component, meaning that plugins can set this to `null` to prevent it being sent in Minecraft's ingame chat. 

DiscordSRV should recognise this way of preventing chat messages being sent about Advancements, and mirror the same behaviour.

### DIFF
--- a/src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
+++ b/src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
@@ -88,9 +88,9 @@ public class PlayerAdvancementDoneListener implements Listener {
 
         // return if message is null as plugins can set this to prevent the advancement being sent in chat
         try {
-            Method message = event.getClass().getMethod("message");
-            if (message.invoke(event) == null) return ;
-        } catch (ReflectiveOperationException e ) {
+            Method messageGetter = event.getClass().getMethod("message");
+            if (messageGetter.invoke(event) == null) return;
+        } catch (ReflectiveOperationException e) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
```diff
@@ -86,6 +86,14 @@ public void onPlayerAdvancementDone(PlayerAdvancementDoneEvent event) {
         // return if advancement or player objects are knackered because this can apparently happen for some reason
         if (event.getAdvancement() == null || player == null) return;
 
+        // return if message is null as plugins can set this to prevent the advancement being sent in chat
+        try {
+            Method message = event.getClass().getMethod("message");
+            if (message.invoke(event) == null) return ;
+        } catch (ReflectiveOperationException e ) {
+            return;
+        }
```
Scarsz, line 95:
```suggestion
        try {
            Method messageGetter = event.getClass().getMethod("message");
            if (messageGetter.invoke(event) == null) return;
        } catch (ReflectiveOperationException e) {
            return;
        }
```

Scarsz (changes requested):

DaMarine:
@Scarsz thanks for the quick review - applied the requested changes ✅

Scarsz:
@DaMarine thinking about this again wouldn't it be better for this to silently ignore the failure if reflection fails? This would currently cause any environment where `message` isn't a valid getter to not function at all. Would be better if it was silently ignored in my opinion

PR #1882, opened 2026-08-08
continue onPlayerAdvancementDone() upon message reflection failure

DaMarine:
Follow-on PR from #1881 

Addresses the later comment that we should silently ignore any reflection failure, and continue with the advancement message in chat as normal.


src/main/java/github/scarsz/discordsrv/listeners/PlayerAdvancementDoneListener.java
```diff
@@ -91,7 +91,7 @@ public void onPlayerAdvancementDone(PlayerAdvancementDoneEvent event) {
             Method messageGetter = event.getClass().getMethod("message");
             if (messageGetter.invoke(event) == null) return;
         } catch (ReflectiveOperationException e) {
-            return;
+            // send advancement in chat as normal upon reflection failure
```
DaMarine, line 94: I'm partial to leaving comments explaining any `catch {}` that doesn't have code within it, for the sake of posterity, but happy to change if you'd prefer otherwise

